PR #53, opened 2014-12-19
Typo in the readme for the browserify version

alexfiedler:
the instruction to npm install gave a package does not exist error.
I changed auth0.js to auth0-js which worked for me.


pose:
@alexfiedler thanks!

PR #64, opened 2015-02-04
Added Recaptcha supports

tehsis:
This expect an error property indicating that recaptcha will be required on next login attempt.

Lock will use it at https://github.com/auth0/lock/pull/118

### DIFF
--- a/lib/LoginError.js
+++ b/lib/LoginError.js
@@ -40,6 +40,7 @@ function LoginError(status, details) {
   err.status = status;
   err.name = obj.code;
   err.code = obj.code;
+  err.require_recaptcha = obj.require_recaptcha;
   err.details = obj;
 
   if (status === 0) {
--- a/test/user-and-pass.tests.js
+++ b/test/user-and-pass.tests.js
@@ -134,6 +134,26 @@ describe('Auth0 - User And Passwords', function () {
         done();
       });
     });
+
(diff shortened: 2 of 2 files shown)


lib/LoginError.js
```diff
@@ -40,6 +40,7 @@ function LoginError(status, details) {
   err.status = status;
   err.name = obj.code;
   err.code = obj.code;
+  err.require_recaptcha = obj.require_recaptcha;
```
pose: Thanks for the contribution. Can this be done using the `code` attribute? Thanks!

jfromaniello:
Let's say that I am using auth0.js to render my own login page. Auth0 returns an error indicating that I need to show the user a captcha:

1.  What does the application needs to render? 
2.  How does the application sends the token to the server?

tehsis:
The client must be able to associate its recaptcha credentials to its Auth0 account.

So, when auth0 requests a recaptcha token, the app must send, in its next login attempt, the token issued by recaptcha and auth0 must verify it (https://developers.google.com/recaptcha/docs/verify)

Obtaining the recaptcha token is responsability of the recaptcha widget: https://developers.google.com/recaptcha/docs/display
You can see my implementation on lock: https://github.com/auth0/lock/pull/118 as an example of the workflow.

PR #76, opened 2015-04-07
Adding detection of connection refused/timeout

### LINKED ISSUE #75: Not properly handeling "Connection Timeout" and "Connection Refused" exceptions
auth0.6.1.0.js is not handling "Connection Timeout" and "Connection Refused" exceptions.  

There was a recent fix to handle "Internet Disconnected" [commit 68aea5dee564c60ede71e0668468f7433f723f3b], but the fix does not cover the situation when the browser can't connect to Auth0 (for example: Auth0 service is down or some network issues).  The error reported is "invalid_password" which may be confusing to users.

pose:
> Fixes #75

Previously, the case where the browser was online and the connection failed was handled on the same manner as when on IE10/11 CORS failed. A new IE version check was added to report the CORS error. When both conditions are true (not IE and a 0 status code) that means that a request to Auth0 failed because of connection refuse or timeout.

### DIFF
--- a/lib/index.js
+++ b/lib/index.js
@@ -83,33 +83,34 @@ function checkIfSet(obj, key) {
 
 function handleRequestError(err, callback) {
   var er = err;
-  var ieVersion = isInternetExplorer();
+  var isAffectedIEVersion = isInternetExplorer() === 10 || isInternetExplorer() === 11;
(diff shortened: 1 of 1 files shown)


pose:
@cristiandouce done

PR #77, opened 2015-04-10
SMS authentication support

ivancevich:


lib/index.js
```diff
@@ -1430,6 +1433,65 @@ Auth0.prototype.getConnections = function (callback) {
 };
 
 /**
+ * Send SMS to do passwordless authentication
+ *
+ * Examples:
```
pose: You can use @example here, thanks!

lib/index.js
```diff
@@ -1430,6 +1433,65 @@ Auth0.prototype.getConnections = function (callback) {
...
+      Authorization: 'Bearer ' + apiToken
+    },
+    data:         {
+      phone_number:   phoneNumber,
+      connection:     'sms',
+      email_verified: false
+    }
+  })
+  .fail(function (err) {
+    try {
```
pose: There are a couple of failure corner cases handled by this function: https://github.com/auth0/auth0.js/blob/master/lib/index.js#L84-L120 . Do you think we can re use that function here?

pose: Also, there's a flag `this._useJSONP` that we use in other methods: https://github.com/auth0/auth0.js/blob/master/lib/index.js#L303

pose:
Apart from the comments, looks good to me :+1:

lib/index.js
cristiandouce: @pose You cant use jsonp with APIv2. :dancers:

pose: Sorry, @cristiandouce is right :). Disregard this comment :dancer:

PR #79, opened 2015-04-24
Fix incorrect base64 decode

wooorm:
Fixes strings containing two or more dashes and/or underscores.

This affected some users, seemingly at random, especially with large tokens!

### DIFF
--- a/lib/base64_url_decode.js
+++ b/lib/base64_url_decode.js
@@ -18,7 +18,7 @@ module.exports = base64_url_decode;
  */
 
 function base64_url_decode(str) {
-  var output = str.replace("-", "+").replace("_", "/");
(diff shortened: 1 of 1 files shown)


jfromaniello:
nice catch

It seems that we have the same problem here:
https://github.com/auth0/jwt-decode/blob/master/lib/base64_url_decode.js#L4

wooorm:
It does! I found loads more by searching GitHub for `base64_url_decode`, weird!

wooorm:
@pose @jfromaniello Any word on this, interest in implementing?

pose:
@wooorm Merged, please check the 6.4.1 in about an hour. Thanks again!

wooorm:
:+1:

PR #87, opened 2015-05-15
Return empty object on parseHash error instead of null

rolodato:
This facilitates checking for errors when calling parseHash.

Before:
```js
var hash = auth0.parseHash();
// if hash is null, hash.error throws an error
if (hash && hash.error) { ... }
else if (hash) { ... }

// or

var hash = auth0.parseHash();
if (hash) {
  if (hash.error) { ... }
  else { ... }
}
```

After:
```js
var hash = auth0.parseHash();
// {}.error returns undefined, no need to check hash twice
if (hash.error) { ... }
else if (hash) { ... }
```

Thoughts?

### DIFF
--- a/index.js
+++ b/index.js
@@ -389,7 +389,7 @@ Auth0.prototype.parseHash = function (hash) {
   }
   if(!hash.match(/access_token/)) {
     // Invalid hash URL
-    return null;
+    return {};
   }
   hash = hash.substr(1).replace(/^\//, '');
   parsed_qs = qs.parse(hash);
--- a/test/tests.js
+++ b/test/tests.js
@@ -351,7 +351,7 @@ describe('Auth0', function () {
     });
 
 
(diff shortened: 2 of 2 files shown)


rolodato:
Sure, I'll take a look :+1:

rolodato:
I just checked all usages in the project, all error checks are done by doing this:

```js
var hash = auth0.parseHash();
if (!hash) { ... }
```

So looks like we're good to go :+1:

jfromaniello:
I don't like anything of what I see here, for me parseHash should be like this

```javascript
var parseHash = function (hash) {
   return qs.parse(hash.replace(/^\#/, ''));
};
```

It seems is doing weird things right now like validating the audience (????)

I don't like this to return null if the hash doesn't contains `access_token` but i don't like it either to return `{}`.

rolodato:
To be reviewed later when revamping parsing methods. This might be a breaking change for users doing things like `if (!hash) // fail fast`.

PR #92, opened 2015-06-05
Allow disable client SDK information from being sent on each request

benschwarz:
This change addresses two concerns: 

* Some clients may prefer to not send SDK version information with authorisation requests, this code allows them to disable the feature. 

* Use headers rather than querystrings to carry the sdk information where possible

### DIFF
--- a/index.js
+++ b/index.js
@@ -155,6 +155,7 @@ function Auth0 (options) {
     facebook: this._phonegapFacebookLogin
   };
   this._useCordovaSocialPlugins = false || options.useCordovaSocialPlugins;
+  this._sendClientInfo = (typeof options.sendSDKClientInfo === 'undefined') ? true : options.sendSDKClientInfo;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -1111,13 +1140,24 @@ Auth0.prototype.loginWithSocialAccessToken = function (options, callback) {
...
 
   var protocol = 'https:';
   var domain = this._domain;
   var endpoint = '/oauth/access_token';
   var url = joinUrl(protocol, domain, endpoint);
+  var headers = {};
+
+  if ( this._sendClientInfo ) {
+    if ( this._useJSONP ) {
+      query['auth0Client'] = this._getClientInfoString();
```
cristiandouce: Why not place the auth0Client info inside _buildAuthorizationParameters() ?

```
function _buildAuthorizationParameters(appendClientInfo [boolean]) {
  var q = {};
  if (appendClientInfo) {
    q.auth0Client = this._getClientInfoString();
  }
}
```

index.js
```diff
@@ -155,6 +155,7 @@ function Auth0 (options) {
     facebook: this._phonegapFacebookLogin
   };
   this._useCordovaSocialPlugins = false || options.useCordovaSocialPlugins;
+  this._sendClientInfo = (typeof options.sendSDKClientInfo === 'undefined') ? true : options.sendSDKClientInfo;
```
cristiandouce:
```
null == null // true
null == undefined // true
null == false // false
null == true // false
```

`this._sendClientInfo = null != options.sendSDKClientInfo ? options.sendSDKClientInfo : true;`

index.js
benschwarz: It'd remove a lot of duplication… good

PR #106, opened 2015-09-01
Fix issue with double opened popups when doing signup

cristiandouce:


pose:
@cristiandouce I reviewed this change :+1:

cristiandouce:
thanks! :smile:

PR #107, opened 2015-09-03
Little refactor for popup opener when auto_login mode on signup

cristiandouce:


index.js
```diff
@@ -497,23 +496,16 @@ Auth0.prototype.signup = function (options, callback) {
 
   var popup;
 
-  if (options.popup && !this._getCallbackOnLocationHash(options)) {
-    popup = this._buildPopupWindow(options);
-  }
+  var will_popup = options.auto_login && options.popup
```
glena: if auto_login = false, it never should show the popup because no sso cookie will be created right? (it wasnt checking this before)

cristiandouce: yes. Exactly :smile:

PR #124, opened 2015-12-28
Changes needed by Lock classic and passwordless

gnandretta:
- Lock passwordless needs to be easy to use Webpack, so we are storing the version number in a regular javascript file.
- Lock classic needs to work with electron, so we are providing a way to override calls to `window.open`.

More details are provided in the commit messages.


cristiandouce:
:+1:

PR #131, opened 2016-02-18
Add support for scopes from options parameters

jsbejeau:
This pull request contains modifications to allow Auth0 Lock to specify multiple scopes for JWT and avoid  to have a fixed scope in the JWT return.


gnandretta:
Hi, the `scope` options is already supported.

PR #139, opened 2016-04-11
Changed three instances of "connection: 'tests' " to "connection: 'Username-Password-Authentication' "

bingdai:
Changed three instances of "connection: 'tests' " to "connection: 'Username-Password-Authentication' ". 

Otherwise the three respective function (auth0.login, auth0.signup, auth0.validate.User) will give either error 400 or error 404. 

Note: I was using the credentials of my own Auth0 application, instead the one provided by this example ("domain: 'mdocs.auth0.com', callbackURL: window.location.href, clientID: '0HP71GSd6PuoRYJ3DXKdiXCUUdGmBbup' ").

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -96,7 +96,7 @@
                 $('.login-db').click(function (e) {
                     e.preventDefault();
                     auth0.login({
-                        connection: 'tests',
+                        connection: 'Username-Password-Authentication',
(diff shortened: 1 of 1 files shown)


gnandretta:
Thanks for reporting this. However, test should be run with the current applications and connections because they rely on their settings.

Issue #168, opened 2016-05-27
Auth0.js crashes phone


calclavia:
I tried using Auth0.js v7.0.2 with an Ionic 2 project. It works in the browser, but every time it is run on mobile, I receive the error "Something went wrong" with no additional information.

I then decided to roll back to previous Auth0.js version v6.8. Installing the app in my phone (Samsung S6) permanently damaged the System UI. The phone now keeps displaying "System UI has stopped" even after restart (must factory reset phone). This is replicable.

Tried:
```
	this.auth.login({
					popup: true,
					connection: this.providerID,
					authParams: {
						scope: "openid name offline_access",
						device: "Mobile device"
					}
				},
				//	(err, result) => {
				(err, profile, id_token, access_token, refreshToken) => {
```

gnandretta:
Hi, could you provide us a sample project that replicates the problem? Also does this happen when installing the application? when starting it? or when attempting to login?

Thank you.

calclavia:
Follow up: So I updated my phone's OS, and it no longer crashes with Auth0 v6.8.

I'll see if I can replicate it on an empty project. I remember this problem happened on two different projects.

cosemansp:
I had the same problem. There is a fundamental bug in the index.js.
You can find a fix in https://github.com/cosemansp/auth0.js (see index.js - startHandler function)
I submitted a pull request

gnandretta:
Thank you @cosemansp. I've added a question in the PR.

gnandretta:
Fixed in #172 and #173

PR #172, opened 2016-06-16
fix issue with 'Something went wrong' error on device (#168)

cosemansp:
See comments in code

### DIFF
--- a/index.js
+++ b/index.js
@@ -840,11 +840,19 @@ Auth0.prototype.loginPhonegap = function (options, callback) {
       return ref.close();
     }
 
-    if (result.id_token) {
-      setTimeout(function() { callback(null, _this._prepareResult(result)) }, 0);
+    // parseHash returns idToken, NOT id_token
+    // if (result.id_token) {
+    //   setTimeout(function() { callback(null, _this._prepareResult(result)) }, 0);
+    //   answered = true;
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -840,12 +840,20 @@ Auth0.prototype.loginPhonegap = function (options, callback) {
...
-      setTimeout(function() { callback(null, _this._prepareResult(result)) }, 0);
+    // parseHash returns idToken, NOT id_token
+    // if (result.id_token) {
+    //   setTimeout(function() { callback(null, _this._prepareResult(result)) }, 0);
+    //   answered = true;
+    //   return ref.close();
+    // }
+
+    if (result.idToken) {
+      callback(null, result);
```
gnandretta: @cosemansp is there any reason to avoid the `setTimeout` here?

cosemansp: No, but I did't had a reason to add it, during the error handling there is also no timeout.

cosemansp: The problem was not in the setTimeout, it was the result.id_token (which was wrong) and the prepareResult was not needed because the tokenPayload was already decoded in the parse function above

gnandretta: If an error happens during the callback execution, the window will remain opened.

gnandretta:
Thanks!

PR #181, opened 2016-08-05
Updated reqwest dependency.

Josecc:
Updates Request dependency to allow for isomorphic javascript rendering.

PR #187, opened 2016-08-18
Adapt code depending on window to support Isomorphic bundling

selaux:
We want to use lock in a isomorphic environment. This requires us to be able to at least require the lock module in nodejs. Currently it is not possible because of several accesses to the `window` variable at require-time. This PR takes care of removing some of these accesses. Other accesses are in:

- reqwest dependency (#181)
- winchan

### DIFF
--- a/lib/json-parse.js
+++ b/lib/json-parse.js
@@ -3,6 +3,6 @@
  * exists on `window`
  */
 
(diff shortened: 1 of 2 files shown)


selaux:
Any thoughts?

PR #201, opened 2016-09-12
Allow to use v2 logout endpoint

hzalaz:
By default uses v1 for backwards compat


index.js
```diff
@@ -1642,15 +1642,34 @@ Auth0.prototype.getDelegationToken = function (options, callback) {
...
  *     // redirects to -> 'https://yourapp.auth0.com/logout?returnTo=http://logout'
  *
+ * @example
+ *
+ *     auth0.logout(null, {version: 'v2'});
+ *     // redirects to -> 'https://yourapp.auth0.com/v2/logout'
+ *
+ * @example
+ *
+ *     auth0.logout({returnTo: 'http://logout'}, {version: 2});
```
glrodasz: You have a typo here. It should be `{version: 'v2'}`.

Issue #170, opened 2016-06-10
reqwest version used expect a window object


kandros:
reqwest version 1.1.4 from package jason expect a window object, I'm building universal application and this is causing a compilation error #127

More recent versions of reqwest does handle the situation in wich window is not defined.

please consider update

phuna:
@kandros : Did you get through this? I'm having the sample problem

kandros:
@phuna i forked and updated reqwest manually, and used https://github.com/webpack/imports-loader when possible

phuna:
@kandros Thanks for the information

MacsDickinson:
I'm getting this too trying to add auth0 to a universal react app.

It's being = looked at here https://github.com/auth0/auth0.js/pull/181

itsovereasy:
I'm suddenly running into this issue as well. No idea why it wasn't happening before, as I've been using this lib and running CI with mocha for a month now.

Winchan is also causing issues due to its calling of the browser global `navigator`

This is causing my builds to fail!

PR #202, opened 2016-09-12
Bump to latest reqwest@2.0.5

cristiandouce:
Fixes https://github.com/auth0/auth0.js/issues/170

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,7 +31,7 @@
     "json-fallback": "0.0.1",
     "jsonp": "~0.0.4",
     "qs": "^6.2.1",
-    "reqwest": "^1.1.4",
+    "reqwest": "2.0.5",
     "trim": "~0.0.1",
     "winchan": "auth0/winchan#v0.1.2",
     "xtend": "~2.1.1"


paulwithap:
Is the Winchan dependency issue also fixed in this? It makes a call to window.navigator that blows up in non browser environments.

hzalaz:
@paulwithap no, as you can see in the changes of the PR it only updates the reqwest depedency

paulwithap:
I'm on my phone and just woke up, so I figured I was missing something ;) Is there a reason that was left out? This only solves half of the problem.

hzalaz:
Because we wanted to fix the issue of the outdated dependency not the isomorphic one (we updated more outdated dependencies). Until we replace winchan it will always be a half fix, and we are planning to rethink the lib to better fit that use case. The current version was not design/implemented with considering server side rendering, this was meant to be a client library used inside a browser.

paulwithap:
Understandable. Do you have any recommendations for how to handle this in testing? I'm not even using server side rendering, but both of these issues suddenly started crashing mocha when I run tests. I'd love to keep using Auth0, but if it means I can't run tests, I obviously can't.

hzalaz:
That's odd, we run our test in mocha for auth0.js, we are using panthomjs and saucelabs (you can check the repo). You should probably split your fronted from your backend tests or mock Auth0.js

paulwithap:
The odd thing is that I was using your lib for weeks before it became an issue, and I can not for the life of me figure out what caused it to start. No amount of Git time travel will fix it. Also, I don't have a backend, so my code is about as isolated as it gets. I'll check out your config and see if I can get it working. This has been driving me nuts!

tomoakley:
@paulwithap did you get anywhere with this? Currently trying to use auth0-js with erikras' [react-redux-universal-hot](https://github.com/erikras/react-redux-universal-hot-example) app but getting the `window is not defined` error from `lib/use_jsonp.js`. Any help would be much appreciated.

glena:
Hi, we are soon to release auth0.js v8 and we are focusing all the effort in that. That version will not have a heavy dependency on the window object. Anyway, this SDK is intended for client side authentication. For your backend, this is the we to go https://github.com/auth0/node-auth0/ or https://auth0.com/docs/quickstart/webapp/nodejs

PR #244, opened 2016-11-22
Added nonce and status to mitigate replay attacks

glena:


glena:
Do not enforce state and nonce when code is used

PR #262, opened 2016-12-07
fail if localstorage is not available to store the autogenerated nonce

glena:


glena:
@smaniotto it is out of scope for v7. It will be reworked in v8.

It really is a responsibility of the developer to handle state and nonce, the idea of this is to try to make easy the happy path

index.js
```diff
@@ -547,8 +547,18 @@ Auth0.prototype.parseHash = function (hash, options) {
...
-    var nonce = options.nonce || window.localStorage.getItem('com.auth0.auth.nonce');
-    window.localStorage.removeItem('com.auth0.auth.nonce');
+    var nonce;
+
+    if (options.nonce) {
+      nonce = options.nonce;
+    } else if (window.localStorage) {
+      try {
+        nonce = window.localStorage.getItem('com.auth0.auth.nonce');
+        window.localStorage.removeItem('com.auth0.auth.nonce');
```
smaniotto: getItem and removeItem don't throw an error if the key is not defined.

smaniotto:
@glena Ok, for both points. Thanks!

PR #263, opened 2016-12-07
Deprecate and warn about getConnections

### LINKED ISSUE #196: Remove the `getConnections` function
Remove the `getConnections` function, as suggested by @siacomuzzi in https://github.com/auth0/api2/issues/1177#issuecomment-241327532

Relevant code: https://github.com/auth0/auth0.js/blob/master/index.js#L1738-L1740

hzalaz:
Closes #196

### DIFF
--- a/index.js
+++ b/index.js
@@ -295,7 +295,7 @@ Auth0.prototype._configureOfflineMode = function(options) {
 
 Auth0.prototype._getUserInfo = function (profile, id_token, callback) {
 
-  console.warn("DEPRECATION NOTICE: This method will be soon deprecated, use `getUserInfo` instead.")
+  warn("DEPRECATION NOTICE: This method will be soon deprecated, use `getUserInfo` instead.")
 
   if (!(profile && !profile.user_id)) {
     return callback(null, profile);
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1913,24 +1913,13 @@ Auth0.prototype.getSSOData = function (withActiveDirectories, cb) {
...
- *       expect(conns[0].name).to.eql('Apprenda.com');
- *       expect(conns[0].strategy).to.eql('adfs');
- *       expect(conns[0].status).to.eql(false);
- *       expect(conns[0].domain).to.eql('Apprenda.com');
- *       expect(conns[0].domain_aliases).to.eql(['Apprenda.com', 'foo.com', 'bar.com']);
- *     });
- *
  * @method getConnections
  * @param {Function} callback
+ * @deprecated This method is deprecated. If you need to get the connections please use Management API https://auth0.com/docs/api/management/v2#!/Connections/get_connections
```
gene1wood: @hzalaz Is there an new auth0.js method that calls the v2 management API (an equivalent to this now deprecated method)?

gene1wood:
@hzalaz What would be a good replacement endpoint to call in the management api v2 that would provide something similar to `getConnections`?

PR #273, opened 2016-12-16
Return policy attr in errors + responseType validation

### LINKED ISSUE #271: [v8] SignupAndLogin blows up if no response type is set
we need to do the validation before the transaction manager does its magic and check other methods. ReponseType should be always mandatory

glena:
Closes #271


src/web-auth/redirect.js
```diff
@@ -36,6 +37,10 @@ Redirect.prototype.login = function (options, cb) {
 
   this.warn.warning('`webauth.redirect.login` will be soon deprecated, use `webauth.login` instead.');
 
+  assert.check(params, { type: 'object', message: 'options parameter is not valid' }, {
```
hzalaz: Can we test this?

PR #289, opened 2017-01-05
Check for correct options.

### LINKED ISSUE #288: Nonce does not match
Using the `8.0.1` version, I always get a `Nonce does not match` error when 
parsing the token returned by `redirect.loginWithCredentials`.

The issue occurs inside a VueJS2 app.

I have an authentication file creating a client:
```
const auth0 = new Auth0.WebAuth({
  domain: process.env.AUTH0_DOMAIN,
  clientID: process.env.AUTH0_CLIENT_ID,
  responseType: 'token',
  redirectUri: window.location.origin + '/'
})

// login
let login = (username, password) => {
  auth0.redirect.loginWithCredentials({
    connection: 'Username-Password-Authentication',
    email: username,
    password: password,
    scope: 'openid email'
  },
  function (err) {
    if (err) console.error('something went wrong: ' + err.message)
  })
}
```
When the application is called back, I try to get read the provided token:
```
let requiresAuth = (to, from, next) => {
  if (!checkAuth()) {
    console.log('auth failed ...')
    let result = auth0.parseHash(window.location.hash, (err, data) => {
      console.log(err || data, window.location.hash)
    })
    console.log('Result:')
    console.dir(result)
    if (result && result.idToken) {
      console.log('Authentication succeeded')
      localStorage.setItem(ID_TOKEN_KEY, result.idToken)
      next({path: '/'})
    } else {
      console.log('Redirect to login...')
      next({path: '/login'})
    }
  } else {
    next()
  }
}
```

The complete app can be found here: https://github.com/vitactiv/vitactiv-admin/tree/version/lock

*EDIT:* Seems that the README has an error. By checking in the code of the transaction Manager, `responseType` should match `id_token` and not `token` as written in the README.md.

yageek:
The responseType is `token` and not `id_token`. Trying to fix #288

### DIFF
--- a/src/web-auth/transaction-manager.js
+++ b/src/web-auth/transaction-manager.js
@@ -16,7 +16,7 @@ TransactionManager.prototype.process = function (options) {
     return options;
   }
 
-  if (options.responseType.indexOf('id_token') !== -1 && !!options.nonce) {
+  if (options.responseType.indexOf('token') !== -1 && !!options.nonce) {
     return options;
   }
 
@@ -24,7 +24,7 @@ TransactionManager.prototype.process = function (options) {
 
   options.state = transaction.state;
 
-  if (options.responseType.indexOf('id_token') !== -1) {
+  if (options.responseType.indexOf('token') !== -1) {
     options.nonce = transaction.nonce;
   }
 


hzalaz:
@yageek thanks for the PR but the check for nonce is only for id_token. We are fixing it by only validating that for OAuth redirect via /authorize

yageek:
Ok. But why does the transaction manager not attach a `nonce` if I did not specify one in the options before calling `loginWithCredentials`?

hzalaz:
because there is a bug we need to fix to avoid generating nonce for that endpoint

PR #291, opened 2017-01-05
Add a leeway option.

### LINKED ISSUE #290: Token was issued in the future
Using the last master (5f6a4af3fa48e885d338f893a8e81033693285b3), I have an issue when I call the `parseHash` function.

I think the issue is related to the `idtoken-verifier`.

Using the app at: https://github.com/vitactiv/vitactiv-admin/tree/version/lock, I have added some log inside the `verifyExpAndIat` method just before the test:

```
  var iatDate = new Date(0);
  iatDate.setUTCSeconds(iat - this.leeway);
  console.log("Now:", now)
  console.log("iatDate:", iatDate)
  if (now < iatDate) {
    return new error.TokenValidationError('The token was issued in the future. ' +
      'Please check your computed clock.');
  }

  return null;
```

After logging, I got the following as output:
```
Now: Thu Jan 05 2017 18:09:15 GMT+0100 (CET)
iatDate: Thu Jan 05 2017 18:09:16 GMT+0100 (CET)
```
 The token is considered as invalid.

yageek:
Add the possibility to specify a leeway.
This helps to fix #290.

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -72,6 +72,7 @@ function WebAuth(options) {
  * @param {String} options.state [OPTIONAL] to verify the response
  * @param {String} options.nonce [OPTIONAL] to verify the id_token
  * @param {String} options.hash [OPTIONAL] the url hash. If not provided it will extract from window.location.hash
+ * @param {Integer} options.leeway [OPTIONAL] the leeway to use for the token validation. Default to 0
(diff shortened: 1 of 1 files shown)


glena:
Closed in favor of #292.

This PR will introduce a breaking change by changing the signature of a public method.

PR #313, opened 2017-01-16
fix to not convert arrays to objects

offroff:
Arrays are converted to objects. That's a bug. This should fix it


glena:
thx @offroff for the contribution but there is another PR targeting this issue. It will be released today.

Closed in favor of #307

PR #316, opened 2017-01-16
Fallback to math.random if there is no crypto support

glena:


test/helper/object.test.js
```diff
@@ -420,7 +420,7 @@ describe('helpers', function () {
       });
     });
 
-    it.only('should not breack the string', function () {
+    it('should not breack the string', function () {
```
luisrudge: break*

PR #335, opened 2017-01-26
Fixed exception in IframeHandler on postMessage(null, ...)

xurei:
Fixes an exception thrown when `parent.postMessage(null, domain);` is called


glena:
Good catch, thx for the contribution

PR #337, opened 2017-01-30
Popup no owp

glena:


src/helper/url.js
```diff
@@ -0,0 +1,11 @@
...
+function extractOrigin(url) {
+  if (!/^https?:\/\//.test(url)) url = window.location.href;
+  var m = /^(https?:\/\/[\-_a-zA-Z\.0-9:]+)/.exec(url);
+  if (m) return m[1];
+  return url;
+}
+
+module.exports = {
+  extractOrigin: extractOrigin
+};
```
cristiandouce: we might want to revisit this once dealing with Electron and chrome extensions (if we intend to support this mode)

cristiandouce:
⚠️ Conflicts should be resolved first.

PR #367, opened 2017-02-26
Deprecation warning: webauth.login → webauth.authorize

dtinth:
> `webauth.redirect.loginWithCredentials` will be soon deprecated, use `webauth.login` instead.

This is the warning when trying to use `webauth.redirect.loginWithCredentials` (which is [still in the official docs](https://auth0.com/docs/libraries/auth0js#webauth-authorize-). However, `webauth.login` is nowhere to be found.

However, 2a6ef047d32fa43b57d011dbd5a707456f80aba6 renamed the `.login()` method to `.authorize()`. This pull request fixes the deprecation warning.

### DIFF
--- a/src/web-auth/redirect.js
+++ b/src/web-auth/redirect.js
@@ -18,7 +18,7 @@ function Redirect(client, options) {
  * @method loginWithCredentials
  * @param {Object} options
  * @param {Function} cb
- * @deprecated `webauth.redirect.loginWithCredentials` will be soon deprecated, use `webauth.login` instead.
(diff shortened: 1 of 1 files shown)


dtinth:
However I don’t know how to send the username and password through `webAuth.authorize()`.

hzalaz:
@dtinth you don't. That will redirect you to your auth0 account login page and there your user can enter their credentials. 
If that is not good enough you can still use the method since it says it will be soon deprecated since we are exploring replacements beyond https://auth0.com/docs/libraries/auth0js#webauth-client-login- that is a simple xhr

hzalaz:
Also thanks for the PR will merge when I am near my mac

PR #382, opened 2017-03-09
Replace querystring implementation with qs module

selaux:
__Rationale__: Writing a proper querystring parse implementation can be hard, so we should leverage existing modules to be more robust. Also there is an issue with the current implementation described in https://github.com/auth0/lock/issues/913.

__Scope__: Replaces all occurences of `helpers/qs.js` with the `qs` module

__Notes__
- I removed [this test](https://github.com/auth0/auth0.js/blob/master/test/helper/popup-handler.test.js#L83), because it was actually testing the querystring implementation
- I had to adapt [this test](https://github.com/auth0/auth0.js/blob/master/test/helper/iframe-handler.test.js#L160) as pluses in urls are decoded as spaces, the new expected value seems more correct to me, but correct me if i am wrong


luisrudge:
Can you share the bundle size difference with this module as a dependency?

PR #384, opened 2017-03-09
Move to CircleCI

hzalaz:


circle.yml
```diff
@@ -0,0 +1,10 @@
+machine:
+  node:
+    version: v6.1.0
```
luisrudge: maybe should use node LTS?

hzalaz: Its in progress not done yet but I need the PR to test it 😄

PR #388, opened 2017-03-14
Correct spelling mistake in web-auth JSDoc resulting in incorrect autocomplete suggestions

Geeman201:
Only cosmetic, but my IDE autocomplete gave an incorrect parameter list.

Thought I'd just fix it rather than make a new bug.


luisrudge:
Thanks for the PR @Geeman201!

PR #399, opened 2017-03-28
Require WinChan on demand to allow this to be used in browserless environments

globetro:
I am trying to use this package in a react-native app so that I can programmatically interact with the auth0 api. Annoyingly, the `winchan` package does an isInternetExplorer check on require (https://github.com/mozilla/winchan/blob/ac4b142c34daa84bbcb5d8663fad19ce6394cb18/winchan.js#L21) which causes issues in environments like react-native where `navigator.userAgent` does not exist.

### DIFF
--- a/src/helper/popup-handler.js
+++ b/src/helper/popup-handler.js
@@ -1,7 +1,5 @@
 /* eslint-disable no-restricted-syntax */
 /* eslint-disable guard-for-in */
-var WinChan = require('winchan');
-
 var windowHandler = require('./window');
 var objectHelper = require('./object');
 var qs = require('qs');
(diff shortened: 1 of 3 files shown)


hzalaz:
Thanks for the PR but this is not meant for running in react-native or any non-browser based environment. Ideally https://github.com/auth0/react-native-auth0 should be the sdk for RN since the use-cases are different and we want to avoid bloating this library to keep it as small as possible

globetro:
@hzalaz Cool, are you guys planning on adding passwordless support to react-native-auth0?

hzalaz:
Yes, we are behind since its not a big priority now but eventually will catch up with the rest of SDKs.

globetro:
@hzalaz Sounds good, it may be something worth prioritizing soon given Google's recent purchase of Fabric (and with it, Digits), who knows if they'll keep that service around. So people may be looking for alternatives soon. Just a thought, thanks!

brunocascio:
@hzalaz We are using this package to build a library (SDK) cross platform, I mean, the same code for web and react-native. As @globetro mentioned, this is the only issue :(

hzalaz:
@brunocascio we have our RN sdk since the impl of each are different. Even if the language is the same the platforms are completely different so we won't be putting any effort to have a universal js sdk

PR #410, opened 2017-04-12
Add disableExpirationCheck to parameters to WebAuth

mattdodge:
There is currently an option in the idtoken-verifier library that allows for disabling the expiration/issued at check of incoming tokens. However, it is preceded by two underscores which indicates (at least to me) that we should not rely on using that. I would like to see this option become part of the documented/versioned API.

The ability to add a `leeway` is nice but does not cover all cases. Especially given the restriction on leeway is no more than one minute. We're not doing any token validation on the client side and I don't particularly like putting the restriction on my users that their clocks be synced. We've received several bug reports of errors because "tokens are issued in the future" even when the user's system clock is synced with NTP. 

This pull request aims to "lock in" the use of the `disableExpirationCheck` parameter so that it doesn't get wiped out in future releases and also doesn't discourage people from using it if they don't want to verify `iat` or `exp` values on id tokens.

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -37,6 +37,7 @@ function WebAuth(options) {
     scope: { optional: true, type: 'string', message: 'scope is not valid' },
     audience: { optional: true, type: 'string', message: 'audience is not valid' },
     leeway: { optional: true, type: 'number', message: 'leeway is not valid' },
+    disableExpirationCheck: { optional: true, type: 'boolean', message: 'disableExpirationCheck is not valid' },
     plugins: { optional: true, type: 'array', message: 'plugins is not valid'},
     _disableDeprecationWarnings: { optional: true, type: 'boolean', message: '_disableDeprecationWarnings option is not valid' },
     _sendTelemetry: { optional: true, type: 'boolean', message: '_sendTelemetry option is not valid' },
@@ -150,7 +151,7 @@ WebAuth.prototype.parseHash = function (options, cb) {
       issuer: this.baseOptions.token_issuer,
       audience: this.baseOptions.clientID,
       leeway: this.baseOptions.leeway || 0,
-      __disableExpirationCheck: this.baseOptions.__disableExpirationCheck
+      __disableExpirationCheck: this.baseOptions.disableExpirationCheck
     });
 
     var decodedToken = verifier.decode(parsedQs.id_token);
@@ -194,7 +195,7 @@ WebAuth.prototype.validateToken = function (token, state, nonce, cb) {
     issuer: this.baseOptions.token_issuer,
     audience: this.baseOptions.clientID,
     leeway: this.baseOptions.leeway || 0,
-    __disableExpirationCheck: this.baseOptions.__disableExpirationCheck
+    __disableExpirationCheck: this.baseOptions.disableExpirationCheck
   });
 
   verifier.verify(token, nonce, function (err, payload) {
--- a/test/helper/iframe-handler.test.js
+++ b/test/helper/iframe-handler.test.js
@@ -88,7 +88,7 @@ describe('helpers iframeHandler', function () {
         clientID: 'gYSNlU4YC4V1YPdqq8zPQcup6rJw1Mbt',
         responseType: 'token',
         _sendTelemetry: false,
-        __disableExpirationCheck: true
+        disableExpirationCheck: true
       });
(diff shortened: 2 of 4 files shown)


luisrudge:
Hi. Thanks for this PR. After some internal discussions, we decided this option is not something we want to allow as first class and it is intended to be hidden so it can be removed in a future release.

mattdodge:
So, just to make sure I'm correct, we should expect all client sessions that use the auth0 library to have their clocks synced within 1 minute of NTP time?

hzalaz:
We can increase the max of the leeway but at one point the customers should have a clock as much up to date as possible. Most OS now sync by default with some ntp server and having a diff of hours or days can be a bigger issue. 

I think an updated max of 15 minutes might be enough (now its a minute)

mattdodge:
I'm curious to why I need my users to have their clocks synced if my application does not require client-side token validation. As it stands, the auth0 library requires this unless I can disable the expiration check.

hzalaz:
So you validate the `id_token` server side by making a request to your server? or you don't rely on it?

mattdodge:
We validate our own tokens on our own server side, yes. If I recall correctly though the token that is causing this validation to fail is the token from the auth0 login. Out of curiosity, why is it "less secure" for a client to have their clock out of sync. I can see if you were generating tokens why it would be an issue, but I can't think of anything nor find any RFC to explain why retrieving a token needs a synced clock unless you wanted to validate the `iat` or `exp` attributes.

hzalaz:
If you validate the token you need to validate signature, and those claims at least (you could also validate that the token is addressed to you) thats why the clock needs to be in sync. The `id_token` is meant only for the client not for the server and the `access_token` is the one meant to be consumed for an API. Validating the `id_token` in the server (not our default recommendation) is fine but you could easily do the same in the client if you use RS-like signing alg. 

I'd rather let you parse the hash yourself since you dont need any extra validation besides checking that the `state` you receive back is the same you sent in the Auth request, than make an option first class for anyone to use without thinking about the consequences.

mattdodge:
I think we've gotten a bit off track from the original issue, my apologies. The main problem this PR was attempting to solve was that our users could not log in to our application if their clock was off because the auth0 library was internally validating the `exp` claim on the `id_token` coming back from the Auth0 servers. Our application does not require any client-side token validation, but auth0.js does. 

This PR allows me to configure the lock widget (or custom auth login widget) to not check that expiration. Is there a reason auth0's client-facing library should not accept a login if the user's clock is off?

hzalaz:
Because they validate the `iat` and if its in the future it will reject the token. That's why I suggested increasing the max leeway to around 15'.

mattdodge:
So back to my initial question, why do we care when an `id_token` was issued when we are logging in? When granting access based on that id token we certainly care, but that is on the server side. If a client gets an identity token why should we validate that? In some cases we may want to, but by rejecting this PR you are saying that in *every* case we should be validating it, that is where I disagree.

hzalaz:
Its a standard validation we make so we only accept valid id_token, and a client should validate it since the id_token is meant for them not for an api to grant access, since it only proves who the user is and some of its information. 
On the other hand, the access_token is the one that proves what access the bearer of it has, and its must be treated as opaque by the client and only validated server side.

This will start to be our recommendation soon, probably I should have started with this explanation. We are in process of updating our docs and providing a migration guide so all of our customers understand the way forward

Back to your setup, my recommendation now will be to parse yourself the hash and validate the state parameter by yourself. 

We won't make that flag public now, its only for backwards compatibility with Lock but once we change  our recommendation we will proceed to phase it out. You are free to use it now just make sure when you update that we haven't removed the flag, I am 80% sure we will issue a new major for auth0.js to clean some methods that are no longer necessary but just in case check that its still there before updating to a new minor.

mattdodge:
Ok, we will hold off on updating the library for now then. Writing my own security parser goes against pretty much all of my intuition haha. Would you entertain a similar pull request to your [idtoken-verifier](https://github.com/auth0/idtoken-verifier) library or would we be having the same conversation there?

hzalaz:
Mostly the same convo sadly. I can make the max leeway a bit more flexible (something we are considering doing already).

The parser is not that hard since its just converting the query string into an object and comparing by value the returned state vs the expected state. This is all you need to do as long as you validate the tokens in the server.

PR #419, opened 2017-05-02
Adding new LoginTicket flow (with session)

luisrudge:


src/web-auth/index.js
```diff
@@ -374,6 +377,43 @@ WebAuth.prototype.signupAndAuthorize = function (options, cb) {
...
+  };
+  var realm = options.realm || this.baseOptions.realm;
+  if (realm) {
+    authenticateBody.realm = realm;
+    authenticateBody.credential_type = 'http://auth0.com/oauth/grant-type/password-realm';
+  }
+  this.request.post(url).withCredentials().send(authenticateBody).end(function (err, data) {
+    if (err) {
+      var redirectUrl = _this.baseOptions.redirectUri || options.redirectUri;
+      var errorHash = '#error=' + err.response.error + '&error_description=' + encodeURI(err.response.error_description);
```
hzalaz: Can response or one of it attributes be undefined?

src/web-auth/index.js
```diff
@@ -374,6 +377,43 @@ WebAuth.prototype.signupAndAuthorize = function (options, cb) {
...
 /**
+ * Logs in the user with username and password. You can use `username` or `email` as the actual username.
+ *
+ * @method login
+ * @param {Object} options options used in the {@link authorize} call after the login_ticket is acquired
+ * @param {String} options.username username
+ * @param {String} options.email email
+ * @param {String} options.password user password
+ * @param {String} options.realm realm
+ * @see   {@link https://auth0.com/docs/api/authentication#login}
```
hzalaz: We don't have a link for this yet so lets remove it

src/web-auth/index.js
```diff
@@ -374,6 +377,43 @@ WebAuth.prototype.signupAndAuthorize = function (options, cb) {
 };
 
 /**
+ * Logs in the user with username and password. You can use `username` or `email` as the actual username.
```
hzalaz: We need to better explain this since its not as simple as the authorize

PR #420, opened 2017-05-03
Fix postMessage handler to handle parsed objects as well

### LINKED ISSUE #417: V8.6 uncaught type error
Upgrading from V8.5 to V8.6 produces the following error which wasn't present previously:
```
Uncaught TypeError: Cannot read property 'replace' of null
    at n.parseHash (auth0.js:3)
    at n.callback (auth0.js:3)
    at n.callbackHandler (auth0.js:3)
    at n.messageEventListener (auth0.js:3)
    at usePostMessage.transientMessageEventListener (auth0.js:3)
```

luisrudge:
fix #417

### DIFF
--- a/example/callback.html
+++ b/example/callback.html
@@ -1,17 +1,8 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <script src="/auth0.js"></script>
     <script type="text/javascript">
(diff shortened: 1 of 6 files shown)


johnlim:
@hzalaz @luisrudge Thanks for the fix guys!

PR #430, opened 2017-05-09
Fixed issues with overrides not being used

sandrinodimattia:
The original code was using `this.overrides` which is never set.


selaux:
Should this be added to the `README`? It seems to be part of the public API and there is no information about what overrides are and what they can be used for...

hzalaz:
@selaux no since its for internal use for now and subject to change. If we have a public one we will document it

PR #432, opened 2017-05-15
Add option to filter iframe events to prevent incorrect events triggering callbacks

### LINKED ISSUE #406: Ignore messages that are not from Auth0's iframe
So renewAuth is not working for me with postMessage enabled. The request is getting cancelled in the browser. In the iframe it looks like other events are being sent (before the code reaches parseHash) which causes the iframe to be destroyed.

Now I have this when running locally (because of an event from webpack) and when running a deployed version (because of an empty event). Here is what fixed it for me in the iframe handler:

```js
IframeHandler.prototype.messageEventListener = function (e) {
  if (!e.data || e.data === '' || e.data.type === 'webpackOk') {
    return;
  }
  this.destroy();
  this.callbackHandler(e.data);
};
```

aaronchilcott:
This change fixes an issue with the Iframe handler, where unrelated events were triggering callbacks, which caused undesired behaviour.

* Adds event validation to the Iframe handler, which allows for customised event filtering.
* Shifts responsibility of what to return in the callback from the Iframe handler to the silent authentication handler.
* Moves some logic out of the Iframe handler and in to the authentication handler, hopefully, more clearly defining the responsibilities of each class.

### DIFF
--- a/example/callback.html
+++ b/example/callback.html
@@ -1,9 +1,12 @@
 <!DOCTYPE html>
 <html>
-  <head>
-    <script type="text/javascript">
-      parent.postMessage(window.location.hash, "https://localhost:3000/");
-    </script>
-  </head>
-  <body></body>
+
+<head>
+  <script type="text/javascript">
+    parent.postMessage(window.location.hash, "https://localhost:3000/");
+  </script>
+</head>
+
+<body></body>
+
 </html>
--- a/example/index.html
+++ b/example/index.html
@@ -345,7 +345,6 @@
       webAuth.renewAuth({
         usePostMessage: true,
         scope:'',
-        audience: 'https://brucke.auth0.com/userinfo',
         redirectURI: 'https://localhost:3000/example/callback.html'
       },
         htmlConsole.dumpCallback.bind(htmlConsole)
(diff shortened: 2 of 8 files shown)


aaronchilcott:
Here's an example of a silent authentication HTML callback file:

    <!DOCTYPE html>
    <html>
    <head>
      <script type="text/javascript">
        var data = {
          type: 'auth0:silent-authentication',
          hash: window.location.hash
        };
    
        // This page is designed to be loaded in an Iframe
        var result = parent.postMessage(data, "*");
      </script>
    </head>
    <body></body>
    </html>

luisrudge:
hey @aaronchilcott! Thanks for the PR. We liked this approach very much. Can you fix the failing tests and add some tests to test this new logic? We want to make sure that this is not a breaking change (which using your default it isn't) and won't break current behaviour for everyone. Let me know if you need some help with that! Again, thank you very much! 🎉

aaronchilcott:
Hi @luisrudge, thanks.
I have updated the test cases. There's an issue with one of the tests failing in the CI environment, however, it does not fail locally for me. Would you please take a look at it?

Thanks,
Aaron

luisrudge:
@aaronchilcott thank you very much for this PR 🎉

aaronchilcott:
@luisrudge, don't mention it. Happy to help 😊

PR #436, opened 2017-05-16
Fix some examples and docs + using https everywhere

luisrudge:


package.json
```diff
@@ -22,7 +22,7 @@
     "test:watch": "mocha --watch -R min test/**/*.test.js",
     "ci:test": "istanbul cover _mocha --report lcovonly -R test/**/* -- -R mocha-multi --reporter-options spec=-,mocha-junit-reporter=-",
     "ci:coverage": "codecov",
-    "lint": "eslint ./src",
+    "lint": "eslint --fix ./src",
```
hzalaz: Please remove the `--fix`

src/web-auth/index.js
```diff
@@ -381,15 +381,15 @@ WebAuth.prototype.signupAndAuthorize = function (options, cb) {
...
+ * Logs in the user with username and password using the cross origin authentication flow. You can use either `username` or `email` to identify the user.
  *
  * @method login
  * @param {Object} options options used in the {@link authorize} call after the login_ticket is acquired
- * @param {String} options.username username
- * @param {String} options.email email
- * @param {String} options.password user password
- * @param {String} options.realm realm
- * @param {Function} cb callback function called only when an authorization error occurs. Has the error as the only parameter.
+ * @param {String} [options.username] Username
```
hzalaz: Explain that `username` and `email` are mutually exclusive here too

hzalaz: And if both are supplied explain what happens if you don't raise an error

src/web-auth/index.js
```diff
@@ -381,15 +381,15 @@ WebAuth.prototype.signupAndAuthorize = function (options, cb) {
...
+ * @param {String} [options.email] Email
+ * @param {String} options.password Password
+ * @param {String} [options.realm] Use the realm paramater to specify a connection. Ex: 'Username-Password-Authentication'
```
hzalaz: `realm used to authenticate the user, it can be a realm name or a database connection name`

src/web-auth/index.js
```diff
@@ -381,15 +381,15 @@ WebAuth.prototype.signupAndAuthorize = function (options, cb) {
...
+ * @param {Function} cb Callback function called only when an authorization error occurs. Has the error as the only parameter
```
hzalaz: Define the callback type as I did in the rest of jsdocs

PR #451, opened 2017-06-07
Fixed issue in overrides of __tenant and __tenant_issuer

dctoon:
The overrides of `__tenant` and `__tenant_issuer` are not mapped correctly when having following config, this PR fixes this.

```
var auth0 = new auth0.WebAuth({
  overrides: {
    __tenant: 'XXX',
    __token_issuer: 'https://XXX.auth0.com/'
  }
});
```

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -79,11 +79,10 @@ function WebAuth(options) {
     : true;
 
   this.baseOptions.tenant =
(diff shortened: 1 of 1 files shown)


dctoon:
Checking the tests, wanted to add a test, but there is one already.

dctoon:
Looks to be fixed in the latest version.

Issue #477, opened 2017-08-01
iOS Safari 10.3.3 SecurityError for localStorage


hwestbrook:
- [x] Code snippet or sample project that reproduces the bug
- [x] Screenshots when appropriate
- [x] Auth0.js version = 8.7.0
- [x] Browser & OS = iOS Safari 10.3.3

Make sure to include **as much information as possible** for us to understand and reproduce the bug, that way we can fix it as quickly as possible.

-----------------------

I'm using Auth0-lock & React for a web app I've created. It works great, except when I have a user the has cookies turned off in iOS. When I have that situation, I'm getting the following error:

![image](https://user-images.githubusercontent.com/335309/28850085-5400384e-76cf-11e7-825c-49342d73a685.png)

I've traced this error back to auth-js, specifically, [this line](https://github.com/auth0/auth0.js/blob/master/src/helper/storage/handler.js#L8): 
```
this.storage = windowHandler.getWindow().localStorage || new CookieStorage();
```

This seems to be a relatively [common issue](https://www.google.com/search?q=SecurityError+%28DOM+Exception+18%29%3A+The+operation+is+insecure.+localstorage&oq=SecurityError+%28DOM+Exception+18%29%3A+The+operation+is+insecure.+localstorage&gs_l=psy-ab.3...1548.5074.0.5178.17.14.2.0.0.0.167.1420.9j5.14.0....0...1.1.64.psy-ab..1.1.104...0.fVbOw3L6YRM).

I've fixed it within node_modules for my project by changing the above to:
```
try {
   this.storage = windowHandler.getWindow().localStorage || new CookieStorage();
} catch(e) {
   this.storage = new CookieStorage();
}
```

luisrudge:
Can you send a PR with this? That'd be amazing. 🎉

hwestbrook:
OK, I've done that

luisrudge:
fixed in https://github.com/auth0/auth0.js/pull/698

PR #478, opened 2017-08-01
fix localStorage security error on ios

hwestbrook:
This fixes https://github.com/auth0/auth0.js/issues/477


luisrudge:
Hi @hwestbrook. Thanks for the PR! Do you mind adding a test to [this](https://github.com/auth0/auth0.js/blob/master/test/helper/storage-handler.test.js) file? I'll merge when it's ready. 🎉

AVVS:
Some notes on the PR based on #570

1. Try to use `sessionStorage` is possible
2. Verify that you can actually use it, because it will throw QuotaError during certain conditions

```js
var storage;
var $w = windowHandler.getWindow();

try {
  storage = window.localStorage;
  storage.setItem('x', '1');
  assert(storage.getItem('x') === '1');
  storage.removeItem('x');
  this.storage = storage;
} catch (e) {
  storage = new CookieStorage();
}

this.storage = storage;
````

Last, but not the least, leverage code that already does that from folks at mozilla, while I understand its a 3rd party dependency, doing all the cases right is not as easy as it sounds ;)

luisrudge:
1: why?
2: this is already implemented in the setItem method (if we can't write, we'll fallback to another storage)

I agree we should use a 3rd party for this. We just don't have the time to do research and migrate right now.

AVVS:
1. since it writes down state & nonce - temporary data - you'd avoid the need to cleanup local storage
2. didnt notice, that should take care of it

luisrudge:
1) I like that, but we use storage to save other types of data as well. Maybe we need to rethink how we handle storage from the ground up, taking in consideration short and long lived objects. We can't do it right now, though :(

luisrudge:
I'm rebuilding the storage layer and this should go away when I get this done.

PR #500, opened 2017-09-01
Adding web_message flow

luisrudge:


test/web-auth/web-auth.test.js
```diff
@@ -1259,4 +1252,181 @@ describe('auth0.WebAuth', function() {
...
+    afterEach(function() {
+      TransactionManager.prototype.process.restore();
+      if (IframeHandler.prototype.init.restore) {
+        IframeHandler.prototype.init.restore();
+      }
+    });
+    it('inits IframeHandler with correct params', function(done) {
+      stub(IframeHandler.prototype, 'init', function() {
+        expect(this.url).to.be(
+          'https://me.auth0.com/authorize?client_id=...&response_type=token&redirect_uri=http%3A%2F%2Fpage.com%2Fcallback&from=transaction-manager&response_mode=web_message&prompt=none'
```
lbalmaceda: query params shouldn't have a specific order. You should be testing URLs with a util class that given an url extracts host, domain, path (array), query and hash params (map).

PR #505, opened 2017-09-18
Removing renewSession and keeping only checkSession

luisrudge:


lbalmaceda:
So if no session is present, the user *must* log in again?

luisrudge:
yes, an error is returned `{error: 'login_required'}`

PR #509, opened 2017-09-20
Fix example of parseHash in README

djmitche:
Passing a string leads to the code looking at `<string>.hash`, seeing that it is undefined, and defaulting to `window.location.hash`.  Which *might* be what you wanted, until it isn't!

Also, why is all of this half-documented in the README rather than pointing to https://auth0.com/docs/libraries/auth0js/v8?  In particular, `options` isn't really fleshed out for any of these.


djmitche:
Worth noting, too, that for this function usePostMessage is documented in the README but not in the site docs.

luisrudge:
Thanks for the PR. We indeed have some fragmentation with our docs and are looking at ways to solving it in the near future.

Issue #508, opened 2017-09-19
Silent Authentication not checking that message is from the correct origin or iframe


phBalance:
Hello,
   I was investigating using silent authentication via renewAuth({ usePostMessage: true }) and started getting gibberish back through the callback's authResult parameter.

auth0-js@8.10.0
chrome & linux

  A little more investigation turned up a few problems with the way you're handling things (and explaining why I was getting gibberish):

1) You're not confirming, in either IframeHandler.prototype.eventListener or SilentAuthenticationHandler.prototype.getEventValidator.isValid, that the origin is correct for the message. Something like eventData.event.origin === 'xxx' is required and rejecting if it's not. I believe this is a security concern (c.f. [MDN's postMessage documentation](https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage))

2) Assuming you have a message from the correct origin, there is no check that the message is from the correct iFrame. In my particular case I have another iframe which posts a message while rewnewAuth's iframe is active. The auth0 handler just says "oh, it's a message it must be for me" which is obviously not true in my case but works for a simplistic demo where there are no other messages being passed around. To figure out if the message is coming from the correct iframe something like [this Stack Overflow answer](https://stackoverflow.com/a/20404180) could be followed in conjunction with origin checking.

3) There is no validation of the message content, once you're certain that it's coming from the right place, before shuffling it off to the callback. I guess this one could be viewed as paranoid, but I'm used to not trusting anything not sure how your library expects to behave.

Thoughts on some test cases:
- For same origin it should be sufficient to create 2 iframes and ensure that they don't respond in the expected order thus reversing the messages.
- For different origin create 2 iframes and ensure that the different origin responds and make sure it's rejected.

  I'd consider helping beyond the kludge up I have in my code to make my situation work, but I'm not certain of how your functionality is used in general or if you have this same problem existing elsewhere in your code base.

luisrudge:
Hi. Have you checked this issue/thread? https://github.com/auth0/auth0.js/issues/406#issuecomment-312738290

We haven't documented this yet, but we have a solution for this specific issue. Can you take a look and see if this solves your issue?

phBalance:
@luisrudge 
While I hadn't seen that thread, using postMessageDataType really only solves point 3 above but can generally be used as a work around for point 2 since you can make the type unique enough to not conflict with anything else. The fact that the code is looking at the contents of the message without checking the origin is still a security hole as I understand it.

luisrudge:
[...]

PR #510, opened 2017-09-24
Add origin and source verification for silent authentication iframe postMessage.

phBalance:
A PR for issue #508. 

- Check that the origin and source are as expected before indicating that a postMessage from an iframe isValid.
- Should only be checked when postMessage is truthy.
- The expected postMessage event origin defaults to the same origin as a receiving window unless renewAuth is passed the postMessageOrigin option. postMessageOrigin should be the DOM origin string for redirectUri.
- Add some tests for the new functionality.
- Some reformatting caused by checkin rules.

The changes are potentially breaking as silent authentication is now more selective about accepting messages. For most situations where the iframe src has the same origin as the receiver, there should be no changes required (i.e. don't need to provide a postMessageOrigin). If your iframe src is not on the same origin (that is different protocol, port, or host) then provide postMessageOrigin.

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -82,9 +82,8 @@ function WebAuth(options) {
   this.baseOptions = options;
   this.baseOptions.plugins = new PluginHandler(this, this.baseOptions.plugins || []);
 
-  this.baseOptions._sendTelemetry = this.baseOptions._sendTelemetry === false
-    ? this.baseOptions._sendTelemetry
-    : true;
+  this.baseOptions._sendTelemetry =
+    this.baseOptions._sendTelemetry === false ? this.baseOptions._sendTelemetry : true;
 
   this.baseOptions._timesToRetryFailedRequests = options._timesToRetryFailedRequests
     ? parseInt(options._timesToRetryFailedRequests, 0)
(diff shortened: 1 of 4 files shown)


luisrudge:
Thanks for the amazing PR! 🎉 
I guess I still don't understand why we can't just use `windowHelper.getWindow().origin` as the origin to check? In what situation you'd use a different origin here?

phBalance:
@luisrudge In most cases that will be sufficient which is why it's the default. It all depends on the origin of the src that  you're providing to the iframe. If you're providing something served from the same origin (protocol, host, and port) then that would be all you need. But it's certainly possible to provide a cross origin src (i.e. serve the redirectUri from another part of your website, a different CDN, etc). I'm not sure how frequently people do things that way, but since it doesn't change control to that new origin in the iframe it's certainly a technological possibility.

If you don't like the extra code that brings in, I can take it out. Taking stuff out is easy ;) I would have preferred not to have it and just figure out the origin from the redirectUri, but there are too many variants that make me expect I'd make a mistake since it's not just regexp.

luisrudge:
cool. I'll test this today and merge it if it's ok. Thanks again! 🎉

luisrudge:
Thanks for the amazing PR, @phBalance 🎉

PR #519, opened 2017-10-16
Feature use authenticate everywhere

luisrudge:


src/web-auth/redirect.js
```diff
@@ -1,74 +1,33 @@
...
- */
-
-/**
- * Performs authentication with username/email and password with a database connection
- *
- * This method is not compatible with API Auth so if you need to fetch API tokens with audience
- * you should use {@link authorize} or {@link login}.
+ * Logs in the user with username and password using the cross origin authentication (/co/authenticate) flow. You can use either `username` or `email` to identify the user, but `username` will take precedence over `email`.
+ * This only works when 3rd party cookies are enabled in the browser. After the /co/authenticate call, you'll have to use the {@link parseHash} function at the `redirectUri` specified in the constructor.
  *
```
aaguiarz: It's not entirely true that it works only when 3rd party cookies are enabled, depends on the browser. Maybe we can link to the Cross Origin docs

src/web-auth/redirect.js
```diff
@@ -1,74 +1,33 @@
...
- * @param {String} [options.scope] scopes to be requested during AuthN. e.g. `openid email`
- * @param {credentialsCallback} cb
+ * @deprecated This method will be released in the next major version. Use `webAuth.login` instead.
+ * @param {Object} options options used in the {@link authorize} call after the login_ticket is acquired
+ * @param {String} [options.username] Username (mutually exclusive with email)
+ * @param {String} [options.email] Email (mutually exclusive with username)
+ * @param {String} options.password Password
+ * @param {String} [options.connection] Connection used to authenticate the user, it can be a realm name or a database connection name
+ * @param {crossOriginLoginCallback} cb Callback function called only when an authentication error, like invalid username or password, occurs. For other types of errors, there will be a redirect to the `redirectUri`.
  */
```
aaguiarz: What about responseType, scope, etc?

Shouldn't we remove the deprecated flag?

luisrudge: Moving forward, we'll end up removing .popup and .redirect, right? So, ideally, let's warn about this being deprecated and remove it in the next major. What do you think?

About other options, we explain that all options that can be sent to `authorize`, can be sent here: 
`@param {Object} options options used in the {@link authorize} call after the login_ticket is acquired`

src/web-auth/redirect.js
luisrudge: updated

PR #528, opened 2017-10-25
Feature popup login with credentials with web message

luisrudge:
This PR makes `webAuth.popup.loginWithCredentials` use  the new Cross Origin Authentication flow. After we get the loginTicket from the `co/authenticate` request, we hit the authorize endpoint in an iframe instead of a popup. This still has the same behavior: you don't get a redirect.


src/web-auth/popup.js
```diff
@@ -187,50 +189,12 @@ Popup.prototype.authorize = function(options, cb) {
...
-    objectHelper.pick(options, ['password', 'connection', 'state', 'scope', '_csrf', 'device'])
-  );
-  params.options.username = options.username || options.email;
-
-  url = urljoin(this.baseOptions.rootUrl, 'sso_dbconnection_popup', options.clientID);
-  relayUrl = urljoin(this.baseOptions.rootUrl, 'relay.html');
-
-  return popup.load(url, relayUrl, { params: params }, responseHandler(cb));
+  options.realm = options.realm || options.connection;
+  delete options.connection;
```
lbalmaceda: Is this intended to be used on legacy mode as well? If that's true, we should keep it.

src/web-auth/redirect.js
```diff
@@ -26,7 +26,7 @@ function Redirect(auth0, options) {
  * @param {crossOriginLoginCallback} cb Callback function called only when an authentication error, like invalid username or password, occurs. For other types of errors, there will be a redirect to the `redirectUri`.
  */
 Redirect.prototype.loginWithCredentials = function(options, cb) {
-  options.realm = options.connection;
+  options.realm = options.realm || options.connection;
   delete options.connection;
```
lbalmaceda: Same comment as before

src/web-auth/web-message-handler.js
```diff
@@ -15,7 +15,10 @@ function runWebMessageFlow(authorizeUrl, options, callback) {
       }
     },
     timeoutCallback: function() {
-      callback({ error: 'timeout', error_description: 'Timeout during fetching SSO data' });
+      callback({
+        error: 'timeout',
+        error_description: 'Timeout during executing web_message communication'
```
lbalmaceda: isn't this a breaking change? I don't see it documented for the users.

src/web-auth/popup.js
```diff
@@ -187,50 +189,12 @@ Popup.prototype.authorize = function(options, cb) {
...
+  options.popup = true;
+  options.responseType = options.responseType || '';
```
lbalmaceda: shouldn't we remove the key rather than sending an empty string?

test/web-auth/popup.test.js
```diff
@@ -486,6 +478,25 @@ describe('auth0.WebAuth.popup', function() {
...
+      var expectedOptions = {
+        email: 'me@example.com',
+        password: '123456',
+        scope: 'openid',
+        realm: 'the_connection',
+        popup: true,
+        responseType: ''
+      };
+      stub(TransactionManager.prototype, 'process', function(options) {
+        delete options.popupHandler;
```
lbalmaceda: Feels like cheating. Rather than removing it I would check that it's actually present.

test/web-auth/popup.test.js
```diff
@@ -486,6 +478,25 @@ describe('auth0.WebAuth.popup', function() {
...
+        expect(options).to.be.eql(expectedOptions);
+        return options;
+      });
+      stub(CrossOriginAuthentication.prototype, 'login', function(options, cb) {
+        delete options.popupHandler;
```
lbalmaceda: Same as above

src/web-auth/web-message-handler.js
luisrudge: error_description is not meant to be parsed by anyone. I just made it more generic, since we're now using web_message for different things.
In any case, the v9 release is a new major, so that's ok.

test/web-auth/popup.test.js
luisrudge: we can't compare in the same way because it has functions and the `expect(...).to.be.eql(..)` doesn't work with functions. In any case, we remove popupHandler in the next method.

src/web-auth/popup.js
luisrudge: the `transactionmanager.process` call expects to have a responseType available. since it's an option parameter, I'm just making sure we don't throw an exception if that's the case.

src/web-auth/popup.js
luisrudge: since we don't know if the user has oidc turned on or not, do you think we should send both?

src/web-auth/popup.js
lbalmaceda: I don't know the internals but what I mean is that `responseType=""` should not be sent to the server. Either send a correct value or not send it at all.

luisrudge: ah, gotcha. I'll fix ti

src/web-auth/web-message-handler.js
lbalmaceda: I'm asking because in some sdks we rely on this "error_description" field to parse (wrong designed) error responses. I'm ok with this change as long as it's documented somewhere soon. i.e. add a file describing the possible configuration/runtime errors.

luisrudge: cool. we'll add to the migration guide.

test/web-auth/popup.test.js
lbalmaceda: then first check if `options.popupHandler` exists and is !== null/undefined, then remove it, and then do the optA.to.be.eql(optB)

luisrudge: what for? I don't care about the popupHandler at all, that's why I'm not asserting anything related to it.

src/web-auth/popup.js
lbalmaceda: Probably legacy mode won't understand the `realm` parameter, and of course for the new one we shouldn't be sending deprecated parameters. Do we have other option than sending both? Can you please try if that works fine on both pipelines?

test/web-auth/popup.test.js
lbalmaceda: I don't know. I thought that checking that is another way of ensuring that you're using the popup mode. But if you think it's worthless then don't. 👍

src/web-auth/redirect.js
luisrudge: this works with both oidc and non-oidc clients.

src/web-auth/transaction-manager.js
```diff
@@ -11,20 +11,21 @@ function TransactionManager(options) {
 
 TransactionManager.prototype.process = function(options) {
   var transaction;
+  var responseType = options.responseType || '';
 
-  if (options.responseType.indexOf('code') !== -1) {
+  if (responseType.indexOf('code') !== -1) {
```
lbalmaceda: This is wrong as it's returning the options without checking if `id_token` is present in the response type. Think of `code id_token`: a`nonce` will be needed anyway.

src/web-auth/transaction-manager.js
```diff
@@ -11,20 +11,21 @@ function TransactionManager(options) {
...
     return options;
   }
 
-  if (options.responseType.indexOf('id_token') !== -1 && !!options.nonce) {
+  if (responseType.indexOf('id_token') !== -1 && !!options.nonce) {
```
lbalmaceda: what will happen if `responseType=id_token` and no nonce set?

src/web-auth/transaction-manager.js
```diff
@@ -11,20 +11,21 @@ function TransactionManager(options) {
...
     return options;
   }
 
   transaction = this.generateTransaction(options.appState, options.state, options.nonce);
 
   options.state = transaction.state;
 
-  if (options.responseType.indexOf('id_token') !== -1) {
+  if (responseType.indexOf('id_token') !== -1) {
     options.nonce = transaction.nonce;
```
lbalmaceda: What happens if the user has defined his own options.nonce?

src/web-auth/transaction-manager.js
luisrudge: return raw options

src/web-auth/transaction-manager.js
luisrudge: it will be already out of this codepath because of https://github.com/auth0/auth0.js/pull/528#discussion-diff-146922975R20

src/web-auth/index.js
```diff
@@ -162,37 +159,54 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
         return cb(validationError);
       }
-      return cb(null, buildParseHashResponse(parsedQs, applicationStatus, payload));
+      return cb(null, buildParseHashResponse(parsedHash, applicationStatus, payload));
     });
   }
 
-  if (parsedQs.id_token) {
+  if (parsedHash.id_token) {
     var verifier = new IdTokenVerifier({
```
lbalmaceda: What's the difference between this validation and the one above? I guess the above is verifying the token's signature while this one ignores it. We shouldn't be ignoring the signature, is there an option?

src/web-auth/index.js
```diff
@@ -162,37 +159,54 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
       issuer: this.baseOptions.token_issuer,
       audience: this.baseOptions.clientID,
       leeway: this.baseOptions.leeway || 0,
       __disableExpirationCheck: this.baseOptions.__disableExpirationCheck
```
lbalmaceda: what's the default value of `baseOptions. __disableExpirationCheck`?

src/web-auth/index.js
```diff
@@ -162,37 +159,54 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
+      return cb(null, buildParseHashResponse(parsedHash, applicationStatus, payload));
     });
   }
 
-  if (parsedQs.id_token) {
+  if (parsedHash.id_token) {
     var verifier = new IdTokenVerifier({
       issuer: this.baseOptions.token_issuer,
       audience: this.baseOptions.clientID,
       leeway: this.baseOptions.leeway || 0,
```
lbalmaceda: I don't see the `jwksCache` key set. If this is doable it's preferred as will save a lot of resources https://github.com/auth0/idtoken-verifier#idtokenverifier

src/web-auth/index.js
```diff
@@ -162,37 +159,54 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
+ * @method validateAuthenticationResponse
+ * @param {Object} options
+ * @param {String} options.hash the url hash. If not provided it will extract from window.location.hash
+ * @param {String} [options.state] value originally sent in `state` parameter to {@link authorize} to mitigate XSRF
+ * @param {String} [options.nonce] value originally sent in `nonce` parameter to {@link authorize} to prevent replay attacks
+ * @param {String} [options._idTokenVerification] makes parseHash perform or skip `id_token` verification. We **strongly** recommend validating the `id_token` yourself if you disable the verification.
+ * @param {authorizeCallback} cb
+ */
+WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash, cb) {
+  var state = parsedHash.state || options.state;
```
lbalmaceda: where are you validating that this `response.state` matches the one stored in `transaction.state`?

src/web-auth/index.js
luisrudge: no default

src/web-auth/index.js
luisrudge: we don't use one. you can add more details here so we can track it: https://github.com/auth0/auth0.js/issues/529

src/web-auth/index.js
luisrudge: not sure why, but auth0js never validated that.

lbalmaceda: we should do it.

src/web-auth/index.js
luisrudge: one only decodes, the other one verifies it.

src/web-auth/index.js
luisrudge: just added state validation

test/plugins/cordova.test.js
```diff
@@ -144,7 +144,7 @@ describe('auth0.plugins.cordova', function() {
           },
           appStatus: null,
           refreshToken: 'kajshdgfkasdjhgfas',
-          state: 'theState',
+          state: null,
```
lbalmaceda: why did you remove the state check on this tests?

test/web-auth/web-auth.test.js
```diff
@@ -410,6 +418,56 @@ describe('auth0.WebAuth', function() {
...
+      var webAuth = new WebAuth({
+        domain: 'mdocs.auth0.com',
+        redirectUri: 'http://example.com/callback',
+        clientID: '0HP71GSd6PuoRYJ3p',
+        responseType: 'token'
+      });
+
+      var data = webAuth.parseHash(
+        {
+          hash: '#state=123&access_token=VjubIMBmpgQ2W2&id_token=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJodHRwczovL21kb2NzLmF1dGgwLmNvbS8iLCJzdWIiOiJhdXRoMHw0QVpERjU2Nzg5IiwiYXVkIjoiMEhQNzFHU2Q2UHVvUllKM0RYS2RpWENVVWRHbUJidXAiLCJleHAiOjE0Nzg1NjIyNTMsImlhdCI6MTQ3ODUyNjI1M30.LELBxWWxcGdYTaE_gpSmlNSdcucqyrhuHQo-s7hTDBA&token_type=Bearer&refresh_token=kajshdgfkasdjhgfas'
```
lbalmaceda: How can be so sure that you won't break this in the future? We need to explicitly tell in this test that the transaction is null so it's clearer to which value you're comparing 123 against.

src/web-auth/index.js
luisrudge: We're absolutely not verifying in some cases. When `options._idTokenVerification` is false, we don't verify it, we only decode it.

src/web-auth/index.js
```diff
@@ -162,37 +159,62 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
+    });
+  }
+  var transactionNonce =
+    options.nonce || (transaction && transaction.nonce) || this.baseOptions.nonce || null;
 
   var applicationStatus = (transaction && transaction.appStatus) || null;
-  if (parsedQs.id_token && options._idTokenVerification) {
-    return this.validateToken(parsedQs.id_token, transactionNonce, function(
+  if (parsedHash.id_token && options._idTokenVerification) {
+    return this.validateToken(parsedHash.id_token, transactionNonce, function(
```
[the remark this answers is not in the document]
luisrudge: it's a public api. I won't break it now.

test/plugins/cordova.test.js
luisrudge: because all the tests assume a certain behavior from transactionManager, which is a hell to test. So I changed them to use the simplest flow possible and  wrote specific tests for the transaction manager itself.

test/web-auth/web-auth.test.js
luisrudge: fixed

src/web-auth/index.js
```diff
@@ -162,37 +159,62 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
+  var transaction = this.transactionManager.getStoredTransaction(state);
+  var transactionStateMatchesState = transaction && transaction.state === state;
+  if (state && !transactionStateMatchesState) {
+    return cb({
+      error: 'invalid_token',
+      errorDescription: '`state` does not match.'
+    });
+  }
+  var transactionNonce =
+    options.nonce || (transaction && transaction.nonce) || this.baseOptions.nonce || null;
```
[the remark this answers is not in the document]
luisrudge: if you explicitly provide a nonce, we check that instead. if not, we check the transaction. if there's no transaction, we check if you passed a nonce when you created the webauth object. if not, we set it to null because that's what the idtoken-verifier expects

src/web-auth/index.js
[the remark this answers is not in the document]
lbalmaceda: I thought this was a new major. Or were you referring to something else? https://github.com/auth0/auth0.js/pull/528#discussion_r146895437

src/web-auth/index.js
```diff
@@ -162,37 +159,62 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
 
   var applicationStatus = (transaction && transaction.appStatus) || null;
-  if (parsedQs.id_token && options._idTokenVerification) {
-    return this.validateToken(parsedQs.id_token, transactionNonce, function(
+  if (parsedHash.id_token && options._idTokenVerification) {
```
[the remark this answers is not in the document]
luisrudge: https://github.com/auth0/auth0.js/blob/master/src/web-auth/index.js#L139

true

src/web-auth/index.js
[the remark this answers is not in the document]
lbalmaceda: We should only be checking against `transaction.nonce`, as that's were we kept the "state of the auth" (state/nonce) in the first place.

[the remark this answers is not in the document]
luisrudge: won't do that right now. tracking here: https://github.com/auth0/auth0.js/issues/530

PR #532, opened 2017-10-26
console warning when using timeout parameter with renewAuth

jeremybull:
When I use the `timeout` parameter with `renewAuth()`, it works as expected, but I also get this bogus warning in the browser console:

> Following parameters are not allowed on the `/authorize` endpoint: [timeout]"

Root cause seems to be that `timeout` was not added to the parameter whitelist when it was implemented by 5b638c8, so I have added it.

### DIFF
--- a/src/helper/parameters-whitelist.js
+++ b/src/helper/parameters-whitelist.js
@@ -28,7 +28,7 @@ var authorizeParams = [
   'owp',
   'device',
   'realm',
-
+  'timeout',
   'protocol',
   '_csrf',
   '_intstate',


luisrudge:
Ouch. sorry this is happening. We need to remove the timeout param from the authorize call, not whitelist it. Can you send a PR with that?

jeremybull:
Oops, sorry for rushing that. I created [another PR](https://github.com/auth0/auth0.js/pull/537). Ran the unit tests this time, so hopefully got it right. (I also noticed in passing a similar problem with `popup.authorize()`, so the new PR fixes that too.)

PR #537, opened 2017-10-27
Remove invalid params from authorize url to eliminate console warnings

jeremybull:
Using the `timeout` parameter with `renewAuth()` produces this warning in the console:
> Following parameters are not allowed on the `/authorize` endpoint: [timeout]

Likewise, using `tenant` and `popupOptions` with `popup.authorize()` produces this warning:
> Following parameters are not allowed on the `/authorize` endpoint: [tenant,popup_options]

These options needed to be blacklisted before the `/authorize` url is built.


jeremybull:
Shoot, this included reformatting I didn't intend to make. Closing and submitting another PR.

PR #538, opened 2017-10-27
 Remove invalid params from authorize url to eliminate console warnings

jeremybull:
Using the `timeout` parameter with `renewAuth()` produces this warning in the console:
> Following parameters are not allowed on the `/authorize` endpoint: [timeout]

Likewise, using `tenant` and `popupOptions` with `popup.authorize()` produces this warning:
> Following parameters are not allowed on the `/authorize` endpoint: [tenant,popup_options]

These options needed to be blacklisted before the `/authorize` url is built.

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -82,9 +82,8 @@ function WebAuth(options) {
   this.baseOptions = options;
   this.baseOptions.plugins = new PluginHandler(this, this.baseOptions.plugins || []);
 
-  this.baseOptions._sendTelemetry = this.baseOptions._sendTelemetry === false
-    ? this.baseOptions._sendTelemetry
-    : true;
+  this.baseOptions._sendTelemetry =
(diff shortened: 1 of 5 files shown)


jeremybull:
@luisrudge is there something I should do to initiate a review on this pr? I really appreciated your quick response on my original attempt (#532): I believe this one addresses your feedback. Thanks!

luisrudge:
Hi, sorry to keep you waiting! Can you add the popupOptions blacklist to the test as well? I think it's the only thing missing

jeremybull:
Sorry @luisrudge I'm not following you. Could you be more specific about which test you want me to change and how? (So far I have not added or enhanced any tests with this PR.)

luisrudge:
Sure thing! Ideally, for each method that you're now blacklisting the timeout/popupOptions/tenant properties, you should have a test that you call the method **with** those properties, and then you validate that the authorize url doesn't contain any of those properties.

luisrudge:
fixed in https://github.com/auth0/auth0.js/pull/611

Issue #412, opened 2017-04-15
WebAuth transaction-manager.js: appState or appStatus?


bragma:
I am not totally sure, but I think there is some misnaming of the "appState/appStatus" parameter between parts of the code expecting a transaction to contain an "appStatus" property (ex: https://github.com/auth0/auth0.js/blob/master/src/web-auth/index.js#L134), but when a transaction is generated the same parameter is called "appState", as an example here: https://github.com/auth0/auth0.js/blob/master/src/web-auth/transaction-manager.js#L40, where it is set in the object, but also here https://github.com/auth0/auth0.js/blob/master/src/web-auth/transaction-manager.js#L23 where the transaction options are processed.

luisrudge:
fixed in https://github.com/auth0/auth0.js/pull/539/commits/75c32a33a2b83299e79419e947ef7f9ae5daa15c

PR #539, opened 2017-10-27
Feature add get sso data

luisrudge:
This PR also removes all the flags that would case id_tokens to not be verified.

### DIFF
--- a/plugins/cordova/popup-handler.js
+++ b/plugins/cordova/popup-handler.js
@@ -8,7 +8,7 @@ function PopupHandler(webAuth) {
   this.options = null;
 }
 
-PopupHandler.prototype.preload = function (options) {
+PopupHandler.prototype.preload = function(options) {
   var _this = this;
   var _window = windowHandler.getWindow();
 
@@ -30,7 +30,7 @@ PopupHandler.prototype.preload = function (options) {
 
   this._current_popup = _window.open(url, '_blank', windowFeatures);
 
-  this._current_popup.kill = function (success) {
+  this._current_popup.kill = function(success) {
     _this._current_popup.success = success;
     this.close();
     _this._current_popup = null;
@@ -39,7 +39,7 @@ PopupHandler.prototype.preload = function (options) {
   return this._current_popup;
 };
 
-PopupHandler.prototype.load = function (url, _, options, cb) {
+PopupHandler.prototype.load = function(url, _, options, cb) {
   var _this = this;
   this.url = url;
   this.options = options;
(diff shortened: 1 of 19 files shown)


plugins/cordova/popup-handler.js
```diff
@@ -107,29 +107,28 @@ PopupHandler.prototype.startHandler = function (event, cb) {
...
+  }
 
   var parts = event.url.split('#');
 
   if (parts.length === 1) {
     return;
   }
 
-  var opts = { hash: parts.pop(), _idTokenVerification: false };
+  var opts = { hash: parts.pop() };
```
aaguiarz: So, we skipping this in Cordova, you know why?

src/web-auth/index.js
```diff
@@ -484,6 +486,9 @@ WebAuth.prototype.authorize = function(options) {
 
   params = this.transactionManager.process(params);
 
+  var connection = params.realm || params.connection;
+  storage.setItem('auth0.ssodata.connection', connection);
+
   windowHelper.redirect(this.client.buildAuthorizeUrl(params));
```
aaguiarz: we are doing this in several places, maybe we could have a class that encapsulates reading/setting the localStorage data for SSO?

plugins/cordova/popup-handler.js
luisrudge: I don't we should probably test that with the beta.

src/web-auth/index.js
luisrudge: done

src/authentication/index.js
```diff
@@ -24,7 +25,7 @@ var DBConnection = require('./db-connection');
  * @param {String} [options.audience] identifier of the resource server who will consume the access token issued after Auth
  * @see {@link https://auth0.com/docs/api/authentication}
  */
-function Authentication(options) {
+function Authentication(auth0, options) {
```
jaredhanson: Is this an internal class?  Why did the `auth0` argument get added?  If its not internal, then that will cause breakages.  If its internal, can we document what's expected?

src/web-auth/index.js
```diff
@@ -190,31 +188,35 @@ WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash,
...
   var applicationStatus = (transaction && transaction.appStatus) || null;
-  if (parsedHash.id_token && options._idTokenVerification) {
-    return this.validateToken(parsedHash.id_token, transactionNonce, function(
-      validationError,
-      payload
-    ) {
-      if (validationError) {
+
+  if (!parsedHash.id_token) {
+    return cb(null, buildParseHashResponse(parsedHash, applicationStatus, null));
```
jaredhanson: What is applicationStatus?  And how is it used when there is no id_token?  I can't quite follow this, and it'd be helpful to write some comments that describe what is happening here and why.

src/authentication/index.js
luisrudge: it's internal. we need the webAuth instance to call the checkSession method.

src/web-auth/index.js
luisrudge: applicationStatus is something that the user can save in the transaction status. When finding the code to show you, I found an issue talking about a bug related to this: https://github.com/auth0/auth0.js/issues/412
So, we store it as `appState`, but query as `appStatus`. So, effectively, no one is able to use this because it's totally broken. I'll fix in a new commit.

luisrudge: fix with a test in https://github.com/auth0/auth0.js/pull/539/commits/75c32a33a2b83299e79419e947ef7f9ae5daa15c

PR #547, opened 2017-11-06
Allow multiple calls to checkSession in parallel

NiallArkEnergy:
Currently if I call checkSession with 2 different audiences the iframe handler for both calls will only get the first one to return

### DIFF
--- a/src/web-auth/web-message-handler.js
+++ b/src/web-auth/web-message-handler.js
@@ -11,7 +11,8 @@ function runWebMessageFlow(authorizeUrl, options, callback) {
     timeout: options.timeout,
     eventValidator: {
       isValid: function(eventData) {
-        return eventData.event.data.type === 'authorization_response';
+        return eventData.event.data.type === 'authorization_response' 
+          && options.state === eventData.event.data.response.state;
       }
     },
     timeoutCallback: function() {
--- a/test/web-auth/web-auth.test.js
+++ b/test/web-auth/web-auth.test.js
@@ -1541,7 +1541,9 @@ describe('auth0.WebAuth', function() {
         function(err, data) {}
       );
     });
-    it('eventValidator validates the event data type is `authorization_response`', function(done) {
+    it('eventValidator validates the event data type is `authorization_response` and the state matches the transaction state', function(
+      done
+    ) {
       stub(IframeHandler.prototype, 'init', function() {
         var getEvent = function(type, state) {
           return { event: { data: { type: type, response: { state: state } } } };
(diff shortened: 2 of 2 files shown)


luisrudge:
Thanks for the PR. Can you add a test that actually tests the state property is the same? Also, what happens if state is empty?

NiallArkEnergy:
If state is not provided the transaction manager will create one (this is my preferred use case). Depending on where the state is provided it can prevent parallel calls. Adding a state to new WebAuth will lock it for all checkSession requests, unless it has state as an option. In that case it is up to the user to call checkSession({  state: 'unique state' }) to ensure that parallel calls work correctly

NiallArkEnergy:
#567

luisrudge:
So, what you're saying is that in order to be able to use checkSession more than once, you should manage the state yourself (use the same for every request). Is that the case? What error do you see when you use checkSession in parallel? In this PR, what happens if options.state is null (as in, we're using the state provided in the constructor for example)

NiallArkEnergy:
I would not handle state. The checkSession method builds its own state and it is best to let it do its own thing.

Definitely do not use the same state for every call. This would cause the same issue. If you insist on maintaining your own state they it should be different for every call, or at least for every call that is currently in flight.

If options.state is null the transaction manager provides one. Or as you said the one used in the WebAuth constructor is used. If there is a state provided in the constructor parallel calls will not work in this case

luisrudge:
I think I get it now.

**Before** this PR:
- If you set state in the constructor, parallel calls won't work.
- If you set a state per `checkSession`, parallel calls won't work because the eventValidator doesn't take state in consideration
- If you don't set a state at all, parallel calls won't work because the eventValidator doesn't take state in consideration

**After** this PR:
- If you set state in the constructor, parallel calls won't work.
- If you set a state per `checkSession`, parallel will work work because the eventValidator is now checking for state as well
- If you don't set a state at all, parallel will work work because the eventValidator is now checking for state as well

Did I get this right this time?

NiallArkEnergy:
Yeah that's exactly it

luisrudge:
Alright, perfect. With that in mind, I'll test this next week and merge it. Thanks for your patience 😝

NiallArkEnergy:
Thank you very much. This was my first pull request to this project and a great experience.

luisrudge:
Thank you for taking the time to write the PR. 🎉

NiallArkEnergy:
Any update on this PR?

luisrudge:
Thanks for the PR! :TAD

PR #561, opened 2017-11-14
Fix options.state in WebAuth.validateAuthenticationResponse

damien-gl:
Since latest `auth0-js` release (8.11.x) there is a breaking change preventing us to login into our application because apparently `parseHash` is not taking into account the `options.state` anymore (we are not using the transaction manager of `auth0-js`).

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -178,7 +178,8 @@ WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash,
   var _this = this;
   var state = parsedHash.state;
   var transaction = this.transactionManager.getStoredTransaction(state);
-  var transactionStateMatchesState = transaction && transaction.state === state;
+  var transactionState = options.state || (transaction && transaction.state) || null;
+  var transactionStateMatchesState = transactionState === state;
   if (state && !transactionStateMatchesState) {
     return cb({
       error: 'invalid_token',
@@ -187,7 +188,7 @@ WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash,
   }
   var transactionNonce = options.nonce || (transaction && transaction.nonce) || null;
 
-  var appState = (transaction && transaction.appState) || null;
+  var appState = options.state || (transaction && transaction.appState) || null;
 
   if (!parsedHash.id_token) {
     return cb(null, buildParseHashResponse(parsedHash, appState, null));


luisrudge:
Can you provide a sample of what wasn't working before?

damien-gl:
Currently in our application we are not using the transaction manager of auth0-js and are managing ourself the state & nonce. Therefore, we are providing the state & nonce values as options to the parseHash function but with the new release it's not taking the `options.state` into account.

Here is the code that we are using:

```
getAuth0().parseHash({ state: state, nonce: nonce }, (err, result) => {

}
```

luisrudge:
Can you provide the full snippet code? How are you creating WebAuth? How are you doing the auth request?

luisrudge:
Does it work if you pass the state in the WebAuth constructor?

luisrudge:
ALso, I want to see how you call the authorize method.

aaguiarz:
@damien-gl can you provide more context of what are you trying to achieve? 

We improved how we handle state in this version, you shouldn't need to manage it yourself, but I assume we are missing something for your use case.

We can add this back, but it will be helpful for us to understand the use case.

damien-gl:
Hi @aaguiarz 

In fact we decided some months ago to manage ourself the nonce & state because at that time `auth0-js` was not properly cleaning the localStorage and we were also facing an other issue with one scenario but I don't recall it.

It seems that `auth0-js` has improved transaction management so we can consider to get ride of our custom transaction management and leverage `auth0-js` feature but definitely this will need some development/testing time on our side.

In the mean time, it could be great to have the breaking change introduced in `8.11.X` fixed which will allow us to still benefit from new `auth0-js` releases until we switch to `auth0-js` transaction management.

You can then introduce this breaking change in a major release (e.g. v9).

damien-gl:
@luisrudge @aaguiarz any update on this?

luisrudge:
@damien-gl can you write a test for this? We agreed on let this in again, but we need a test to over this use case. Thanks 🎉

damien-gl:
@luisrudge sorry for the delay, I was quite busy these days.
So I added a unit test to cover the fix.
Can you please have a look quite quickly as we are waiting this fix to upgrade to version 8.12 to have the fix for the following security vulnerability https://auth0.com/docs/security/bulletins/cve-2017-17068
Thanks

luisrudge:
Thanks for the PR 🎉  Keep in mind this is the master branch, which will be released with the new major version 9.0.0.

luisrudge:
If you want this in the 8.x version, can you open a PR to the v8 branch with those changes. I'll approve and release a patch fix right after.

damien-gl:
thanks @luisrudge PR for v8 branch is there:
https://github.com/auth0/auth0.js/pull/591

luisrudge:
Thanks @damien-gl. version 8.12.1 was published with this fix 🎉

PR #563, opened 2017-11-14
Fix iframe bug sometimes occuring in chrome

njonatha:
In certain situations, chrome is not triggering fetch of iframe. This fix simply moves the iframe src update to after it is appended to the DOM.

See auth0 ticket: #31402

### DIFF
--- a/src/helper/iframe-handler.js
+++ b/src/helper/iframe-handler.js
@@ -29,7 +29,6 @@ IframeHandler.prototype.init = function() {
 
   this.iframe = _window.document.createElement('iframe');
   this.iframe.style.display = 'none';
-  this.iframe.src = this.url;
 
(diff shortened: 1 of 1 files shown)


luisrudge:
Can you describe what "certain situations" the bug happen?

njonatha:
I created a reproducer that I attached to the auth0 ticket. The reproducer is a react-app created by create-react-app, and includes the intercom widget code as well (it doesn't reproduce without including intercom too). It also only reproduces in Chrome on Windows. 

The problem occurs whenever pressing "back" from an external site and returning to the app.

Are you able to view the ticket?

saltukalakus:
@luisrudge there is more context and a sample application in ZD # 31402

saltukalakus:
@luisrudge I will DM you. Thanks.

PR #598, opened 2017-12-21
Support browsers that return undefined on window.origin

thoean:
Some browsers don't support window.origin. See [compatibility chart on MDN](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/origin#Browser_compatibility).

This pull requests constructs window.origin if it's not provided.

### DIFF
--- a/src/web-auth/silent-authentication-handler.js
+++ b/src/web-auth/silent-authentication-handler.js
@@ -11,8 +11,7 @@ function SilentAuthenticationHandler(options) {
   this.postMessageOrigin =
     options.postMessageOrigin ||
     windowHelper.getWindow().origin ||
-    windowHelper.getWindow().location.protocol + '//' + windowHelper.getWindow().location.hostname
-      + (windowHelper.getWindow().location.port ? windowHelper.getWindow().location.port : '');
+    windowHelper.getWindow().location.origin;
 }
 
 SilentAuthenticationHandler.create = function(options) {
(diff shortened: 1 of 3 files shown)


luisrudge:
IE11 supports window.origin. Are you sure Edge doesn't support it?

thoean:
Good point. Now I see. IE11 and Edge are the same:
* `window.origin` is `undefined`
* `window.location.origin` matches "scheme + host + port"

I think I should be able to simplify this PR, but I still think some change is needed.

luisrudge:
Cool! I'll keep an eye on the updates. Thanks 🎉

luisrudge:
do all browsers support window.location.origin? if that's the case, I think we're safe to use only that one.

luisrudge:
IE, Edge and Chrome support window.location.origin. Can you test on firefox and safari?

thoean:
Without further testing, it looks like window.location.origin is widely supported. See https://www.w3schools.com/jsref/prop_loc_origin.asp

However, IE only from v11. What are the aspirations regarding compatibility of this library? I could still leave the fallback for manually constructing the origin, my original version.

For sure we can remove window.origin.

luisrudge:
Yeah, so let's keep it like this: 
1: options.postMessageOrigin
2: window.location.origin
3: fallback

Please add a test for each one (I believe we already have a test for 1) and we'll merge it. Thanks 🎉

thoean:
I've just done that. Thank you.

This will merge into v8. Is there anything to do to get this also into the latest / future version?

luisrudge:
Thanks @thoean for the PR and tests 🎉

luisrudge:
Can you open another PR for the master branch? I'll cut new 8.x and 9.x releases today 🎉

thoean:
See https://github.com/auth0/auth0.js/pull/627. I just cherry-picked the merge commit.

PR #602, opened 2017-12-26
setting getSSOData timeout to 5s

luisrudge:


luisrudge:
This is useful for when there's an error in the checkSession call (like invalid web_origin etc). Instead of using the default 60s timeout, it uses 5.

PR #611, opened 2017-12-29
Blacklisting invalid params in authorize url

### LINKED ISSUE #606: Remove popup_options warning
It would be good to not have this warning:

image

while still using `popupOptions` to change the height of the popup:

image

luisrudge:
fix #606

### DIFF
--- a/test/authentication/authentication.test.js
+++ b/test/authentication/authentication.test.js
@@ -64,6 +64,17 @@ describe('auth0.authentication', function() {
       });
     });
 
+    ['username', 'popupOptions', 'domain', 'tenant', 'timeout'].forEach(function(param) {
+      it.only('should remove parameter: ' + param, function() {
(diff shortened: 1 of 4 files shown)


test/authentication/authentication.test.js
```diff
@@ -64,6 +64,17 @@ describe('auth0.authentication', function() {
       });
     });
 
+    ['username', 'popupOptions', 'domain', 'tenant', 'timeout'].forEach(function(param) {
+      it.only('should remove parameter: ' + param, function() {
```
lbalmaceda: isn't this `only` making the test script only run this case?

luisrudge: FUUUUUUU

lbalmaceda: <3

PR #615, opened 2017-12-29
Validate current window origin and redirecturi origin to prevent mismatches

luisrudge:


src/helper/window.js
```diff
@@ -10,8 +10,19 @@ function getWindow() {
   return global.window;
 }
 
+function getOrigin() {
+  var location = global.window.location;
+  var origin = location.origin;
+  if (!origin) {
+    origin =
+      location.protocol + '//' + location.hostname + (location.port ? ':' + location.port : '');
```
cocojoe: Why the new line?

src/web-auth/index.js
```diff
@@ -387,7 +388,7 @@ WebAuth.prototype.checkSession = function(options, cb) {
     .with(options);
 
   if (params.responseType === 'code') {
-    return cb(new Error("responseType can't be `code`"));
+    return cb({ error: 'error', error_description: "responseType can't be `code`" });
```
cocojoe: Why has this changed?

src/web-auth/web-message-handler.js
```diff
@@ -11,8 +12,10 @@ function runWebMessageFlow(authorizeUrl, options, callback) {
     timeout: options.timeout,
     eventValidator: {
       isValid: function(eventData) {
-        return eventData.event.data.type === 'authorization_response' 
-          && options.state === eventData.event.data.response.state;
+        return (
```
cocojoe: Why change?

src/helper/window.js
luisrudge: [prettier](https://github.com/prettier/prettier) auto formatting.

src/web-auth/index.js
luisrudge: it was wrong before. you couldn't catch the error because it was not using the standard error formatting `{error, error_description}`

src/web-auth/web-message-handler.js
luisrudge: also prettier

src/web-auth/index.js
cocojoe: ok good was wondering that

PR #626, opened 2018-01-12
Update v8 README.md - clarify token required for management client

jeffreylees:


luisrudge:
we'll use access tokens instead

PR #627, opened 2018-01-15
Use window.location.origin instead of window.origin

thoean:
This is a repeat of https://github.com/auth0/auth0.js/pull/598 (which was against v8 instead of master). Basically I just cherry-picked 722497e752e7075d9d37abf35ba7a6e67470d44d.


luisrudge:
Thanks very much @thoean 🎉

PR #633, opened 2018-01-22
Fix auth0.authentication docs and readme

luisrudge:


OlivierGonthier:
I lost some time while migrating to v9 because of this.
The first parameter actually seems to be unused in code, and from my testing, passing null doesn't cause any issues. So, should we pass webAuth, or something else, or is this parameter a mistake?

luisrudge:
It's used only for the `getSSOData` method. If you use getSSOData, you need to pass an instance of webAuth.

luisrudge:
This is not needed anymore. You can pass just the options and we'll figure it out how to get the webauth instance.

PR #661, opened 2018-02-09
Fix federated param

luisrudge:


src/authentication/index.js
```diff
@@ -195,7 +195,15 @@ Authentication.prototype.buildLogoutUrl = function(options) {
 
   params = objectHelper.toSnakeCase(params, ['auth0Client', 'returnTo']);
 
-  qString = qs.stringify(params);
+  qString = qs.stringify(objectHelper.blacklist(params, ['federated']));
+  if (
+    options &&
+    (options.federated !== undefined &&
```
joshcanhelp: @luisrudge - Parens not needed here.

Logic is confusing here but I understand why. Just breaking it down ...

* No `options` === no `federated` option so that makes sense
* For the last one, wouldn't `options.federated === 'true'` be a little more explicit here?
* For the 2nd and 3rd, you're checking for `undefined` and `false` in the case where `federated` is in the URL but not set. Wouldn't it just be one or the other or are there browser differences being accounted for? I'm wondering if it's not better to say `Object.keys(options).indexOf('federated') > -1`, then decide the values that should be true (probably 'true' and undefined)

PR #664, opened 2018-02-14
Use webAuth.login when calling signupAndLogin to support Univeral Login Page

luisrudge:
IN order to make `webauth.client.signupAndLogin` support the HLP, we need to make it sure that we call the `webauth.login` after the signup, since this method will pick the right endpoint to login

### DIFF
--- a/src/web-auth/redirect.js
+++ b/src/web-auth/redirect.js
@@ -2,6 +2,7 @@ var CrossOriginAuthentication = require('./cross-origin-authentication');
 var Warn = require('../helper/warn');
 
 function Redirect(auth0, options) {
+  this.webAuth = auth0;
   this.baseOptions = options;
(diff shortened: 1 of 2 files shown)


src/web-auth/redirect.js
```diff
@@ -2,6 +2,7 @@ var CrossOriginAuthentication = require('./cross-origin-authentication');
 var Warn = require('../helper/warn');
 
 function Redirect(auth0, options) {
+  this.webAuth = auth0;
   this.baseOptions = options;
   this.client = auth0.client;
```
lbalmaceda: could this line be removed now, since `this.webAuth.client` is the same as `this.client`?

luisrudge: absolutely. I removed in the next commit

Issue #651, opened 2018-02-05
getOriginFromUrl: IE11 always includes a port


proglottis:
https://github.com/auth0/auth0.js/blob/2bc36781d3f25182e3dfababb79d7335064df73b/src/helper/object.js#L129

In IE11 this seems to always return the port even when it isn't required resulting in this error:
> The redirectUri's origin (https://test.domain.com:443) should match the window's origin (https://test.domain.com).

maxpaj:
Experiencing the same issue.

luisrudge:
I'll try to fix this in the next week. If any of you want to try it out something and submit a PR, I can review it ASAP. Thanks for the patience 🎉

PR #669, opened 2018-02-16
Use WinChan on popup.callback again + adding origin check to keep it secure

luisrudge:
fix https://github.com/auth0/auth0.js/issues/651

### DIFF
--- a/src/helper/object.js
+++ b/src/helper/object.js
@@ -4,7 +4,6 @@
 
 var assert = require('./assert');
 var objectAssign = require('./object-assign');
-var windowHelper = require('./window');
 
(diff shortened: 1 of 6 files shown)


src/helper/object.js
```diff
@@ -118,16 +117,32 @@ function toCamelCase(object, exceptions) {
   }, {});
 }
 
+function getLocationFromUrl(href) {
```
radekk: Seems like this function might require to add a couple of test cases to verify that it's working fine.

luisrudge: added a few of them

radekk: 👍

PR #673, opened 2018-02-20
Throw error when hash.state is empty

luisrudge:


test/web-auth/transaction-manager.test.js
```diff
@@ -133,4 +133,26 @@ context('TransactionManager', function() {
       });
     });
   });
+  context('getStoredTransaction', function() {
+    beforeEach(function() {
+      stub(storage, 'removeItem');
+      stub(storage, 'getItem', function(state) {
```
lbalmaceda: `state` is the "key" value on which you stored the data?

luisrudge: in this specific test, it's anything.. any string at all.. in the context of the library, it's the state, yes.

PR #685, opened 2018-03-02
Adding legacy error handling for co/auth endpoint

luisrudge:


test/web-auth/cross-origin-authentication.test.js
```diff
@@ -166,7 +166,9 @@ describe('auth0.WebAuth.crossOriginAuthentication', function() {
               error_description: 'a huge error string'
             },
             code: 'any error',
```
joshcanhelp: This should probably be a machine_name without spaces

test/web-auth/cross-origin-authentication.test.js
```diff
@@ -166,7 +166,9 @@ describe('auth0.WebAuth.crossOriginAuthentication', function() {
               error_description: 'a huge error string'
             },
             code: 'any error',
-            description: 'a huge error string'
+            description: 'a huge error string',
+            error: 'any error',
```
joshcanhelp: This should probably be a machine_name without spaces

src/helper/response-handler.js
```diff
@@ -42,7 +42,10 @@ function wrapCallback(cb, options) {
       if (err.err) {
         err = err.err;
       }
-
+      if (options.forceLegacyError) {
+        errObj.error = err.error;
```
joshcanhelp: This means that it's possible for `errObj.code` !== `errObj.error` ... is that intended? Maybe this should fall after where `err.code` and `errObj.description` are set and set them as equal?

test/web-auth/cross-origin-authentication.test.js
luisrudge: doesn't really matter for the tests though..

src/helper/response-handler.js
luisrudge: makes sense. pushed a change

PR #688, opened 2018-03-02
Add audience:/userinfo to getSSOData checkSession call

luisrudge:
This prevents auth0-server to return an error when the profile is too big to be returned. This also doesn't affect the end result, considering we only need the `email` and `name` parameters.


cocojoe:
Does this address a report issue or was it just internally reported?

cocojoe:
This will be fixed in server now, so this needed?

luisrudge:
Yeah. This will be handled in the server.

PR #689, opened 2018-03-02
Add __enableImpersonation flag to enable impersonation again

### LINKED ISSUE #683: cannot impersonate (sign in as)
when "sign in as" is used from user management, it triggers the application configured callback:
http://localhost:9000/auth0-callback#access_token=some-access-token&id_token=some-id-token&scope=openid&expires_in=86400&token_type=Bearer
notice, there is no state.

but in auth0.js v9 we have:
https://github.com/auth0/auth0.js/blob/2ca7ccf3ba8e736f859b1024a8327ab0f5f8b110/src/web-auth/index.js#L180-L191

it always trips "state does not match"

luisrudge:
fix #683

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -183,7 +183,7 @@ WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash,
   var transaction = this.transactionManager.getStoredTransaction(state);
   var transactionState = options.state || (transaction && transaction.state) || null;
   var transactionStateMatchesState = transactionState === state;
-  if (!state || !transactionStateMatchesState) {
+  if (!options.__disableStateCheck && (!state || !transactionStateMatchesState)) {
     return cb({
       error: 'invalid_token',
       errorDescription: '`state` does not match.'
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -182,8 +182,11 @@ WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash,
   var state = parsedHash.state;
   var transaction = this.transactionManager.getStoredTransaction(state);
   var transactionState = options.state || (transaction && transaction.state) || null;
+
   var transactionStateMatchesState = transactionState === state;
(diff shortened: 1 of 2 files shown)


src/web-auth/index.js
```diff
@@ -182,8 +182,11 @@ WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash,
   var state = parsedHash.state;
   var transaction = this.transactionManager.getStoredTransaction(state);
   var transactionState = options.state || (transaction && transaction.state) || null;
+
   var transactionStateMatchesState = transactionState === state;
-  if (!state || !transactionStateMatchesState) {
+  var shouldBypassStateChecking = !state && !transactionState && options.__enableImpersonation;
```
lbalmaceda: `state` and `transactionState` are missing for *every* impersonation call? 
Also I don't know how this boolean checks works on JS but probably `options. __enableImpersonation` will be false 99% of the cases, so I'd move that check to the beginning of the line.

src/web-auth/index.js
```diff
@@ -182,8 +182,11 @@ WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash,
   var state = parsedHash.state;
   var transaction = this.transactionManager.getStoredTransaction(state);
   var transactionState = options.state || (transaction && transaction.state) || null;
+
   var transactionStateMatchesState = transactionState === state;
```
lbalmaceda: Won't this be `true` for both values of `null`? e.g. `null === null`  -> `transactionStateMatchesState=true`?

src/web-auth/index.js
luisrudge: state might be present if you used the advanced fields in the page: 
image

`var transactionState = options.state || (transaction && transaction.state) || null;`
transactionState will be either provided by the developer (`options.state`) or empty when you're trying to impersonate

lbalmaceda: So if the user provides a value in that field (image you shared). Where does that end up being added, options.state (transaction state) or the request state?

Issue #691, opened 2018-03-03
login() with email causes /authorize error


SteveALee:
Auth0.js 9.3.2
Firefox  Dev edition on latest Windows
Code snippet using [auth0.js.login() here](https://github.com/AlwaysInMind/aim-web-app/blob/master/client/src/drivers/auth.js#L58-L69) 

Browser console warning  - 'Following parameters are not allowed on the `/authorize` endpoint: [email]'

But login appears to work fine

luisrudge:
are you calling this inside the hosted login page?

luisrudge:
I guess you aren't, since this is part of your app. I'm not sure what's going on: [This](https://github.com/auth0/auth0.js/blob/master/example/index.html#L250) code doesn't throw the warning.

luisrudge:
ahh, sorry. you're using `email` and we're not filtering that out 😁

luisrudge:
The fix will be out in the next release

SteveALee:
Problem seems to have gone. Was this the CDN hosted script changing?

luisrudge:
what cdn link are you using? we didn't release anything

SteveALee:
Please ignore - it *is* still happening

I was looking at 9.3.2 diffs trying to figure out what had changed and appeared to show a cdn lookup. My bad

PR #692, opened 2018-03-03
Remove `email` param in cross auth login

luisrudge:
fix https://github.com/auth0/auth0.js/issues/691


src/web-auth/cross-origin-authentication.js
```diff
@@ -43,9 +43,12 @@ CrossOriginAuthentication.prototype.login = function(options, cb) {
   var _this = this;
   var theWindow = windowHelper.getWindow();
   var url = urljoin(this.baseOptions.rootUrl, '/co/authenticate');
+  options.username = options.username || options.email;
+  delete options.email;
```
cocojoe: Why move the code up here? Just so it `email` can be deleted?

luisrudge: yes

PR #696, opened 2018-03-05
Fix issue #1235 - origin check and sending origin param

lexaurin:
Fix 2) for https://github.com/auth0/lock/issues/1235

- use origin instead of redirectUrl for origin check in postMessage @see https://html.spec.whatwg.org/multipage/web-messaging.html#dom-window-postmessage and https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage
- Fix `Uncaught TypeError: Cannot read property 'protocol' of null` for redirectUrl with protocol other than http(s). RFC https://tools.ietf.org/html/rfc3986 cc. @GabrielArakaki

### DIFF
--- a/src/helper/object.js
+++ b/src/helper/object.js
@@ -119,7 +119,8 @@ function toCamelCase(object, exceptions) {
 
 function getLocationFromUrl(href) {
   var match = href.match(
-    /^(https?:)\/\/(([^:/?#]*)(?::([0-9]+))?)([/]{0,1}[^?#]*)(\?[^#]*|)(#.*|)$/
+    // @see https://tools.ietf.org/html/rfc3986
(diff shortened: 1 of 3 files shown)


luisrudge:
Lock v11 is not going to work in Cordova apps, so you will need to migrate to the Hosted Login Page. More info here: https://github.com/auth0/lock/issues/1235#issuecomment-374740654

vicrep:
@luisrudge I get that you guys won't support cordova, but this affects all applications served from the `file:` protocol. Would it not be reasonable for Lock/auth0-js to support all valid web origins, since this `null` origin behaviour with the protocol is standard?

luisrudge:
If you're using file://, you're probably developing a native app (cordova, electron, ionic etc), which requires [Code Grant with PKCE] (https://auth0.com/docs/api-auth/tutorials/authorization-code-grant-pkce). Lock is not meant to be used with that Grant. Moving forward, we want to make sure  all environments work with Auth0's Universal Login Page, so, if anything you're trying to do with that regard is not working, please reach out to our amazing [support team](https://support.auth0.com).

PR #698, opened 2018-03-08
Use CookieStorage when localStorage is not available

luisrudge:
some browsers throw an error when trying to access localStorage when localStorage is not available

### DIFF
--- a/example/index.html
+++ b/example/index.html
@@ -201,7 +201,7 @@
 
     var webAuth = new auth0.WebAuth({
       domain: 'brucke.auth0.com',
-      redirectUri: 'https://localhost:3000/example',
+      redirectUri: 'https://localhost:3000/example/',
       clientID: 'k5u3o2fiAA8XweXEEX604KCwCjzjtMU6',
       responseType: 'token',
       plugins: [
@@ -211,7 +211,7 @@
 
     var webAuthPasswordless = new auth0.WebAuth({
       domain: 'brucke.auth0.com',
-      redirectUri: 'https://localhost:3000/example',
+      redirectUri: 'https://localhost:3000/example/',
       clientID: 'VZUvmnj3pd9yNgq8BjX4YA8Km14jQ0PN',
       responseType: 'token'
     });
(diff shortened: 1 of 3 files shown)


src/helper/storage/handler.js
```diff
@@ -5,7 +5,14 @@ var Warn = require('../warn');
 
 function StorageHandler() {
   this.warn = new Warn({});
-  this.storage = windowHandler.getWindow().localStorage || new CookieStorage();
```
cocojoe: The PR name is use cookiestorage when local not available. It already did this though? This is a fix to catch errors no?

luisrudge: renamed

example/index.html
```diff
@@ -201,7 +201,7 @@ <h2>Console:</h2>
 
     var webAuth = new auth0.WebAuth({
       domain: 'brucke.auth0.com',
-      redirectUri: 'https://localhost:3000/example',
+      redirectUri: 'https://localhost:3000/example/',
```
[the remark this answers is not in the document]
luisrudge: I was trying it in safari and it doesn't accept the `#access_token` if you don't have a `/` in the end 🤦‍♂️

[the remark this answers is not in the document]
joshcanhelp: Have you tested in other browsers to make sure this addition doesn't do something silly with this addition?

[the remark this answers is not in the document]
luisrudge: chrome/safari/firefox. here's the thing: we already have code that handles when you can't read/write from the storage, the issue was that safari specifically was throwing an exception even before that (just by referencing localSTorage).

src/helper/storage/handler.js
```diff
@@ -5,7 +5,14 @@ var Warn = require('../warn');
...
+  this.storage = new CookieStorage();
+  try {
+    // some browsers throw an error when trying to access localStorage
+    // when localStorage is disabled.
+    this.storage = windowHandler.getWindow().localStorage;
+  } catch (e) {
+    this.warn.warning("Can't use localStorage. Using CookieStorage instead.");
```
lbalmaceda: Maybe you want to add the `e` to the log?

luisrudge: done

PR #706, opened 2018-03-13
Use response_type=token in checkSession when response_type is set to code

luisrudge:


src/web-auth/index.js
```diff
@@ -394,7 +394,7 @@ WebAuth.prototype.checkSession = function(options, cb) {
     .with(options);
 
   if (params.responseType === 'code') {
-    return cb({ error: 'error', error_description: "responseType can't be `code`" });
```
cocojoe: So we used to tell the developer they were implementing it incorrectly and they should fix, but now we just change this internally to `token`, why?

luisrudge: checkSession will never work code, because there's no redirect etc so, instead of throwing an error, it's better to actually use the correct response_type and make it work by default.

lbalmaceda: I agree with @cocojoe here. We shouldn't be fixing incorrect configurations for the users. Instead, tell them how to fix it. Maybe the change should be improving the error message so they can find where to set the "response type" parameter.

luisrudge:
This was fixed by the backend team

PR #707, opened 2018-03-13
Check if the hash contains the tokens requested in the response_type

luisrudge:


src/web-auth/index.js
```diff
@@ -160,6 +160,31 @@ WebAuth.prototype.parseHash = function(options, cb) {
   ) {
     return cb(null, null);
   }
+  var responseType = (this.baseOptions.responseType || options.responseType || '').split(' ');
```
luisrudge: because responseType is not a required field for `WebAuth` (since you can add responseType to other methods individually) and was also never required for `parseHash` as well, I'm making this check optional. It will only run the check when you set a responseType in either the `WebAuth` constructor or `parseHash` method.

cocojoe: There is no default?

luisrudge: not in auth0.js

lbalmaceda: `response_type` is a required field anyway on the API, so even if "other methods can set it", shouldn't you check that it was set? 
source ->https://auth0.com/docs/api/authentication#social

luisrudge: we have this in our docs: 
image


Because of this, people might not be adding responseType to the constructor call, only to the auth request calls.
Since `responseType` was never an option in the `parseHash` method,  we have to deal with the fact that maybe responseType is not defined.

src/web-auth/index.js
```diff
@@ -160,6 +161,31 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
+  if (
+    responseType.length > 0 &&
```
lbalmaceda: If you already have a passing indexOf check over the responseType string var, this one is useless.

test/web-auth/web-auth.test.js
```diff
@@ -365,22 +365,6 @@ describe('auth0.WebAuth', function() {
             };
           });
         });
-        it('sets ssodata with a connection and without a sub when there is no payload', function(
```
lbalmaceda: why this was removed?

test/web-auth/web-auth.test.js
```diff
@@ -801,6 +785,123 @@ describe('auth0.WebAuth', function() {
           }
         );
       });
+      describe('should throw invalid_token error', function() {
```
lbalmaceda: I guess you meant

```
should throw invalid_hash error
```

src/web-auth/index.js
lbalmaceda: Let's go back.
- API Docs say it's required when you call the endpoints.
- Auth0JS docs say it's required as well, since "if you don't specify it here you must specify it on each method".

In which case will Auth0JS work "without a response type being set anywhere" and parseHash be called to obtain a "valid output"?

luisrudge: the scenario we're handling is:

```js
//responseType is not required here
var auth0 = new auth0.WebAuth({
  domain: "{YOUR_AUTH0_DOMAIN}",
  clientID: "{YOUR_AUTH0_CLIENT_ID}"
});

//if you don't use a responseType here, we throw an error
auth0.authorize({responseType: 'token'})

//inside this method, responseType will be undefined
auth0.parseHash(function(err, authResult) {
  console.log(err, authResult);
});
```

test/web-auth/web-auth.test.js
luisrudge: Not sure.. I put it back and added the code changes

src/web-auth/index.js
```diff
@@ -160,6 +161,31 @@ WebAuth.prototype.parseHash = function(options, cb) {
...
+    responseType.indexOf('token') !== -1 &&
```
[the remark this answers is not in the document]
lbalmaceda: Ah ignore me. I miss the call to `split(" ")`. So this is `Array.prototype.indexOf()` and it's basically like calling `if (array contains "token")`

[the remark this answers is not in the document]
luisrudge: yes, that was the reason I used an array instead, because `id_token` always contains `token`

src/web-auth/index.js
lbalmaceda: Even if this is an array, the check is already being done with indexOf. There's no need to add this size check (is redundant)

luisrudge: I renamed to responseTypes to make it more clear

src/web-auth/index.js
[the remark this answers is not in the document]
luisrudge: I renamed to responseTypes to make it more clear

src/web-auth/index.js
lbalmaceda: Thanks! that's super clear. 
So we're adding this check ONLY when a `response_type` is set on the `auth0` object, correct? I'm OK.

luisrudge: either the WebAuth constructor or the parseHash method.

luisrudge: I added a new jsdoc entry in the parseHash method for responseType: https://github.com/auth0/auth0.js/pull/707/files#diff-2f11667c8f671044416e38e4ceb9d7c8R126

PR #708, opened 2018-03-15
Add flag __enableIdPInitiatedLogin to enable idp initiated logins

luisrudge:
this will also enable impersonation, since it's an IdP initiated login as well. The __enableImpersonation flag is still enabled so we don't break anyone that is already using it.

### DIFF
--- a/src/web-auth/index.js
+++ b/src/web-auth/index.js
@@ -179,12 +179,14 @@ WebAuth.prototype.parseHash = function(options, cb) {
  */
 WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash, cb) {
   var _this = this;
+  options.__enableIdPInitiatedLogin =
+    options.__enableIdPInitiatedLogin || options.__enableImpersonation;
   var state = parsedHash.state;
   var transaction = this.transactionManager.getStoredTransaction(state);
   var transactionState = options.state || (transaction && transaction.state) || null;
 
   var transactionStateMatchesState = transactionState === state;
-  var shouldBypassStateChecking = !state && !transactionState && options.__enableImpersonation;
+  var shouldBypassStateChecking = !state && !transactionState && options.__enableIdPInitiatedLogin;
 
   if (!shouldBypassStateChecking && !transactionStateMatchesState) {
     return cb({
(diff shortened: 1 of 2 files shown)


src/web-auth/index.js
```diff
@@ -179,12 +179,14 @@ WebAuth.prototype.parseHash = function(options, cb) {
  */
 WebAuth.prototype.validateAuthenticationResponse = function(options, parsedHash, cb) {
   var _this = this;
+  options.__enableIdPInitiatedLogin =
```
joshcanhelp: Minor but ... I wonder if you shouldn't call this something else since it encompasses 2 different use cases here? Like `options.__bypassStateChecking` or thereabouts?

luisrudge:
We'll have proper docs for this flag shortly: https://github.com/auth0/docs/pull/5857

jamesacres:
var shouldBypassStateChecking = !state && !transactionState && options.__enableIdPInitiatedLogin;

I don't think this solves the issue in our usecase where we set state serverside outside of our auth.js and redirect to authorize. We need a way of bypassing state checking entirely with an option for this case I think?

luisrudge:
if you redirect to `authorize` with the state, `authorize` will redirect to your `redirect_url` with the state as well. You just need to make sure to call parseHash with the state you had in the server `webAuth.parseHash({hash: window.location.hash, state: stateFromTheServer })`

jamesacres:
Thanks for pointing me in the right direction, I have added the following which allows the state check to always pass.

For security maybe we need a way of finding out if a transaction is in progress and only do this if it isn't, or passing another param to our callback page? Otherwise I guess this simply bypasses the check in all cases even if it did originate from auth0.js?

```
                var hash = (window.location.hash) ? window.location.hash : '#' + location.replace(/http.?:\/\/[^/]+/,'').slice(1);

                var parseHashOptions = { 
                    hash:   hash
                };  

                var parsedQs = qs.parse(hash.replace(/^#?\/?/, ''));
                if (parsedQs.hasOwnProperty('state')) {
                    parseHashOptions.state = parsedQs.state;
                }   

                auth0.parseHash(parseHashOptions, authorizeCallback);
```

luisrudge:
Yeah, but then you, as a developer, are intentionally bypassing the state checking, there's nothing we can do. In order to be actually secure, `stateFromTheServer` has to come from the server.

PR #709, opened 2018-03-15
Adding access_token validation for RS256 id_token's

luisrudge:


src/helper/error.js
```diff
@@ -5,11 +5,11 @@ function buildResponse(error, description) {
   };
 }
 
-function invalidJwt(description) {
+function invalidToken(description) {
```
lbalmaceda: can this be considered a breaking change?

test/web-auth/web-auth.test.js
```diff
@@ -307,6 +317,75 @@ describe('auth0.WebAuth', function() {
         }
       ); // eslint-disable-line
     });
+    it('should return the id_token payload when there is no access_token', function(done) {
```
lbalmaceda: shouldn't the payload also be returned when there's a VALID access token? That's the missing test case IMO, even if it's 95% similar to this

test/web-auth/web-auth.test.js
```diff
@@ -307,6 +317,120 @@ describe('auth0.WebAuth', function() {
...
+            state: 'foo',
+            expiresIn: null,
+            tokenType: 'Bearer',
+            scope: null
+          });
+          done();
+        }
+      ); // eslint-disable-line
+    });
+    it('should return the id_token payload when there is an access_token', function(done) {
```
lbalmaceda: when there is a _valid_ access_token

test/web-auth/web-auth.test.js
```diff
@@ -307,6 +317,120 @@ describe('auth0.WebAuth', function() {
...
+            state: 'foo',
+            expiresIn: null,
+            tokenType: 'Bearer',
+            scope: null
+          });
+          done();
+        }
+      ); // eslint-disable-line
+    });
+    it('should validate an access_token when available', function(done) {
```
lbalmaceda: I don't know if you are using the same convention somewhere, but I'd say:

```
should throw a token validation error when the access_token is available but not valid
```

luisrudge: image

PR #717, opened 2018-03-22
Add jwksURI override option

luisrudge:
ZD:37454


src/web-auth/index.js
```diff
@@ -75,8 +75,13 @@ function WebAuth(options) {
       options.overrides,
       { type: 'object', message: 'overrides option is not valid' },
       {
-        __tenant: { type: 'string', message: '__tenant option is required' },
-        __token_issuer: { type: 'string', message: '__token_issuer option is required' }
+        __tenant: { optional: true, type: 'string', message: '__tenant option is required' },
+        __token_issuer: {
```
lbalmaceda: have you run the beautifier here? the others are one liners while this one is split in 3 lines.

luisrudge: nope. I'm not testing idtoken-verifier internals here.

lbalmaceda: > I'm not testing idtoken-verifier internals here.

https://github.com/auth0/auth0.js/pull/717/files/46bc569a1c1920d3f169c49ddbe2acbba9a5d165#diff-548558e7a13abfcf8869090467ec6967R2435

This looks internal to me. So if you're already doing it to check if a value gets replaced, do the same to check that a value remains as expected 👍

luisrudge: idtoken-verifier doesn't set the url in the constructor, it builds later on. I added a test to make sure it's undefined if you don't overwrite.

lbalmaceda: Then in that case both tests should ignore the `idtoken-verifier` implementation and test the actual outcome:

- Have a test that succeeds to verify a token using a *custom* `jwks_uri`.
- Have a test that succeeds to verify a token using the *default* `jwks_uri` (`${id_token.iss}/.well-known/jwks.json`).

Probably you can manage to feed a jwks file to the library. If you don't test this this way, how can you be sure that the `__jwks_uri` override is actually affecting the library verifying outcome?

PR #728, opened 2018-04-03
Overrides for authorize and userInfo Urls

dkhunt27:


luisrudge:
What's the reason for this?

PR #729, opened 2018-04-03
Resolve transaction state to undefined

Neaox:
This fixes an issue where the response state is undefined.


Neaox:
This is not the right for the issue for which it was created. The correct fix for the issue is in https://github.com/auth0/auth0.js/pull/731

PR #733, opened 2018-04-04
getSSOData should call /ssodata from the ULP

luisrudge:
currently, lock [does this](https://github.com/auth0/lock/blob/master/src/core/web_api/p2_api.js#L120) already, but auth0.js will always use checkSession, which is not supported in the ULP. This PR adds compatibility to the ULP when you want to call getssodata.

### DIFF
--- a/src/web-auth/hosted-pages.js
+++ b/src/web-auth/hosted-pages.js
@@ -1,4 +1,9 @@
+var urljoin = require('url-join');
+var qs = require('qs');
+
 var UsernamePassword = require('./username-password');
+var RequestBuilder = require('../helper/request-builder');
+var responseHandler = require('../helper/response-handler');
 var objectHelper = require('../helper/object');
 var windowHelper = require('../helper/window');
 var Warn = require('../helper/warn');
@@ -7,6 +12,7 @@ var assert = require('../helper/assert');
 function HostedPages(client, options) {
   this.baseOptions = options;
   this.client = client;
+  this.request = new RequestBuilder(this.baseOptions);
 
   this.warn = new Warn({
     disableWarnings: !!options._disableDeprecationWarnings
@@ -93,6 +99,35 @@ HostedPages.prototype.signupAndLogin = function(options, cb) {
     }
     return _this.login(options, cb);
   });
+};
+
+HostedPages.prototype.getSSOData = function(withActiveDirectories, cb) {
+  var url;
+  var params = '';
+
+  if (typeof withActiveDirectories === 'function') {
+    cb = withActiveDirectories;
+    withActiveDirectories = false;
+  }
+
+  assert.check(withActiveDirectories, {
+    type: 'boolean',
+    message: 'withActiveDirectories parameter is not valid'
+  });
+  assert.check(cb, { type: 'function', message: 'cb parameter is not valid' });
+
+  if (withActiveDirectories) {
+    params =
+      '?' +
+      qs.stringify({
+        ldaps: 1,
+        client_id: this.baseOptions.clientID
+      });
+  }
+
+  url = urljoin(this.baseOptions.rootUrl, 'user', 'ssodata', params);
+
+  return this.request.get(url, { noHeaders: true }).withCredentials().end(responseHandler(cb));
 };
 
 module.exports = HostedPages;
--- a/test/authentication/authentication.test.js
+++ b/test/authentication/authentication.test.js
@@ -8,6 +8,7 @@ var RequestMock = require('../mock/request-mock');
 var request = require('superagent');
 
 var RequestBuilder = require('../../src/helper/request-builder');
+var windowHelper = require('../../src/helper/window');
 var storage = require('../../src/helper/storage');
 var Authentication = require('../../src/authentication');
 var WebAuth = require('../../src/web-auth');
@@ -17,7 +18,10 @@ var telemetryInfo = new RequestBuilder({}).getTelemetryData();
 describe('auth0.authentication', function() {
   before(function() {
     this.webAuthSpy = {
-      checkSession: spy()
+      checkSession: spy(),
+      _universalLogin: {
+        getSSOData: spy()
+      }
     };
   });
   describe('initialization', function() {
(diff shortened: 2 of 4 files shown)


src/web-auth/hosted-pages.js
```diff
@@ -95,4 +101,33 @@ HostedPages.prototype.signupAndLogin = function(options, cb) {
...
+      '?' +
+      qs.stringify({
+        ldaps: 1,
+        client_id: this.baseOptions.clientID
+      });
+  }
+
+  url = urljoin(this.baseOptions.rootUrl, 'user', 'ssodata', params);
+
+  return this.request.get(url, { noHeaders: true }).withCredentials().end(responseHandler(cb));
```
lbalmaceda: why `noHeaders=true`?

test/authentication/authentication.test.js
```diff
@@ -267,135 +271,175 @@ describe('auth0.authentication', function() {
...
+
+        this.webAuthSpy.checkSession.lastCall.args[1]({
           error: 'consent_required',
-          error_description: 'Consent required. When using `getSSOData`, the user has to be authenticated with the following scope: `openid profile email`.'
+          error_description: 'foobar'
         });
-        expect(result).to.be.eql({ sso: false });
-        done();
       });
+      it('returns ssoData object with lastUsedConnection and idtokenpayload.name when there is no idtokenpayload.email', function(
```
lbalmaceda: idtokenpayload -> idTokenPayload

test/authentication/authentication.test.js
```diff
@@ -267,135 +271,175 @@ describe('auth0.authentication', function() {
...
-          lastUsedUsername: 'last-used-user-name',
-          lastUsedClientID: '...',
-          sessionClients: ['...'],
-          sso: true
+        this.webAuthSpy.checkSession.lastCall.args[1](null, {
+          idTokenPayload: { sub: 'the-user-id', name: 'last-used-user-name' }
         });
-        done();
       });
+      it('returns ssoData object with lastUsedConnection and idtokenpayload.email by default', function(
```
lbalmaceda: idtokenpayload -> idTokenPayload

test/authentication/authentication.test.js
```diff
@@ -267,135 +271,175 @@ describe('auth0.authentication', function() {
...
+      });
+      afterEach(function() {
+        windowHelper.getWindow.restore();
+      });
+      it('calls webauth._universalLogin.getSSOData with same params', function() {
+        this.auth0.getSSOData('withActiveDirectories', 'cb');
+        expect(this.webAuthSpy._universalLogin.getSSOData.lastCall.args).to.be.eql([
+          'withActiveDirectories',
+          'cb'
+        ]);
```
lbalmaceda: missing `done()` call?

test/authentication/authentication.test.js
```diff
@@ -267,135 +271,175 @@ describe('auth0.authentication', function() {
...
-      this.auth0.getSSOData(function(err, result) {
-        expect(err).to.be(null);
-        expect(result).to.be.eql({
-          lastUsedConnection: { name: 'lastUsedConnection' },
-          lastUsedUserID: 'the-user-id',
-          lastUsedUsername: 'last-used-user-email',
-          lastUsedClientID: '...',
-          sessionClients: ['...'],
-          sso: true
+    context('when inside of the hosted login page', function() {
```
lbalmaceda: Since this is a different context, please separate it with 1 or 2 new lines.

test/authentication/authentication.test.js
```diff
@@ -267,135 +271,175 @@ describe('auth0.authentication', function() {
...
-        idTokenPayload: { sub: 'the-user-id', email: 'last-used-user-email', name: 'do not use me' }
+      beforeEach(function() {
+        stub(windowHelper, 'getWindow', function() {
+          return { location: { host: 'me.auth0.com' } };
+        });
+      });
+      afterEach(function() {
+        windowHelper.getWindow.restore();
+      });
+      it('calls webauth._universalLogin.getSSOData with same params', function() {
```
lbalmaceda: Please explain me this test case. Why is it handling the same parameters to the other class's method and why is it accepting strings when it should be a boolean and a function?

test/web-auth/hosted-pages.test.js
```diff
@@ -386,4 +386,67 @@ describe('auth0.WebAuth._universalLogin', function() {
...
+        responseType: 'code',
+        _sendTelemetry: false
+      });
+    });
+
+    afterEach(function() {
+      request.get.restore();
+    });
+
+    it('should call ssodata with all the options', function(done) {
```
lbalmaceda: ssodata -> ssoData

test/web-auth/hosted-pages.test.js
```diff
@@ -386,4 +386,67 @@ describe('auth0.WebAuth._universalLogin', function() {
...
+
+      this.auth0._universalLogin.getSSOData(function(err, data) {
+        expect(err).to.be(null);
+        expect(data).to.eql({
+          sso: false
+        });
+        done();
+      });
+    });
+    it('should call ssodata with all the ad options', function(done) {
```
lbalmaceda: ssodata -> ssoData. Also I'd call it: "should call ssoData with all the options including activeDirectories/ad"

test/web-auth/hosted-pages.test.js
```diff
@@ -386,4 +386,67 @@ describe('auth0.WebAuth._universalLogin', function() {
...
+
+    it('should call ssodata with all the options', function(done) {
+      stub(request, 'get', function(url) {
+        expect(url).to.be('https://me.auth0.com/user/ssodata/');
+        return new RequestMock({
+          headers: {},
+          cb: function(cb) {
+            cb(null, {
+              body: {
+                sso: false
```
lbalmaceda: Can you please add one more parameter to this body? Same for the test below. Why? It makes some noise to me that the `sso: false` is also passed on many of the tests above of this PR, and I want to fully understand what this `getSSOData` call ends doing.

src/web-auth/hosted-pages.js
luisrudge: just copied from v8: https://github.com/auth0/auth0.js/blob/v8/src/authentication/index.js#L382

src/web-auth/hosted-pages.js
```diff
@@ -95,4 +101,33 @@ HostedPages.prototype.signupAndLogin = function(options, cb) {
...
+    type: 'boolean',
+    message: 'withActiveDirectories parameter is not valid'
+  });
+  assert.check(cb, { type: 'function', message: 'cb parameter is not valid' });
+
+  if (withActiveDirectories) {
+    params =
+      '?' +
+      qs.stringify({
+        ldaps: 1,
```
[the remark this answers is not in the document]
luisrudge: just copied from v8: https://github.com/auth0/auth0.js/blob/v8/src/authentication/index.js#L375

src/web-auth/hosted-pages.js
```diff
@@ -95,4 +101,33 @@ HostedPages.prototype.signupAndLogin = function(options, cb) {
   });
 };
 
+HostedPages.prototype.getSSOData = function(withActiveDirectories, cb) {
+  var url;
+  var params = '';
+
+  if (typeof withActiveDirectories === 'function') {
```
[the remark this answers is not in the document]
luisrudge: it has to be the same signature as v8: https://github.com/auth0/auth0.js/blob/v8/src/authentication/index.js#L356

also, callbacks should always be the last argument

test/authentication/authentication.test.js
luisrudge: fixed

test/authentication/authentication.test.js
luisrudge: fixed

test/authentication/authentication.test.js
luisrudge: only needed for async tests

test/authentication/authentication.test.js
luisrudge: fixed

test/web-auth/hosted-pages.test.js
luisrudge: changed to /user/ssodata

test/web-auth/hosted-pages.test.js
luisrudge: sso:false is simulating the server response.

test/web-auth/hosted-pages.test.js
luisrudge: I made AD uppercase

luisrudge:
> I see you're still refering to HLP instead of ULP in the code. Should this be changed? (var names, test descriptions, etc)

I agree with this, but I'll do it in a different PR (replace everything related to HLP to ULP)

test/authentication/authentication.test.js
luisrudge: in the ULP, we need to run the exact same code we were using in v8, so I'm just calling the old getSSOData (now inside the HostedPages class) with the same params. I don't really care what are the params now, just want to test that I'm just relaying the params correctly

src/web-auth/hosted-pages.js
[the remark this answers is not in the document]
lbalmaceda: but what does passing this parameter do?

src/web-auth/hosted-pages.js
lbalmaceda: but what does passing this parameter do?

test/web-auth/hosted-pages.test.js
lbalmaceda: Then why is it called "with all the options" if you're just passing 1? The previous tests I mentioned have more parameters. Is this correct with a single one?

src/web-auth/hosted-pages.js
luisrudge: doesn't send the headers https://github.com/auth0/auth0.js/blob/master/src/helper/request-builder.js#L79

src/web-auth/hosted-pages.js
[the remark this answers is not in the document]
luisrudge: I have no idea what the ssodata endpoint does with this param

test/web-auth/hosted-pages.test.js
luisrudge: yeah.. I just copied/pasted from v8. I renamed the test

src/web-auth/hosted-pages.js
lbalmaceda: This is also removing the telemetry header.. What's the reason for not sending any headers now? I know you copied this from another project, but I need a reason not to add it now.

luisrudge: probably because we don't want to track ssodata? I don't know.

lbalmaceda: @cocojoe after the discussion we had today regarding tracking, is this something you want to track?

Issue #744, opened 2018-04-26
TypeError: Cannot read property 'localStorage' of undefined


it-coral:
I installed auth0-js ( version 9.5.0 ) in node application.
Node version is 6.9.0.

I got the issue after launching application with npm start.

**TypeError: Cannot read property 'localStorage' of undefined**
**....**
**Can't use localStorage. Using CookieStorage instead.**

How can I resolve it?

phlogisticfugu:
also ran into this issue.

we're loading a universal js app (nuxt.js) and its happening when importing the 'auth0-js' library.  The reason is that the auth0-js module is itself trying to instantiate a StorageHandler, and there is no window object on the server.

think that this should be refactored:
https://github.com/auth0/auth0.js/blob/dcf71380ce780490d34d12dcdffed7e858aa11f9/src/helper/storage.js#L2
to just define functions, but not actually run the code that depends upon the window object

phlogisticfugu:
@luisrudge actually the issue is a little more subtle than that

Re: `make sure to only instantiate Auth0.js objects when you're running in the browser`

In my case, we are only running the actual call to `auth0js.WebAuth()` in the browser, and not server side.  This error is being triggered not by that line, but instead in our library import: `import auth0js from 'auth0-js'`

In the particular universal js framework we are using (nuxt.js) imports are done as part of pre-processing bundling, and so the same libraries are loaded on the browser and client.  Loading a module should not result in the execution of code, rather, executing the functions defined in a module should execute code.

see https://github.com/auth0/auth0.js/blob/dcf71380ce780490d34d12dcdffed7e858aa11f9/src/helper/storage.js#L2

which is running code upon module load.

phlogisticfugu:
also note that this error only started appearing in version 9.5.0 of auth0-js

phlogisticfugu:
@luisrudge I made https://github.com/auth0/auth0.js/pull/745 to fix this.

luisrudge:
@phlogisticfugu you're totally right. I'll fix this in a different PR, since we don't want to have that lazy thing again, but thanks for figuring out the issue!

PR #745, opened 2018-04-27
restore lazy instantiation of StorageHandler()

phlogisticfugu:
to avoid instantiating the StorageHandler, which requires the global window object, upon module load

bug was introduced in https://github.com/auth0/auth0.js/pull/737

addresses issue: https://github.com/auth0/auth0.js/issues/744


luisrudge:
I'll fix this in another way. Thanks for the PR!

Issue #402, opened 2017-03-28
localStorage is spammed with state/nonce combos


csvn:
Every time before `auth0.js` performs a redirect to the `/authorize` endpoint, `state` and `nonce` are generated and stored in `localStorage` for verification upon return. If the user lands on the Auth0 hosted login page but does not login, the stored `state/nonce` combo won't be removed.

This means that `localStorage` can fill up with lots of unused values, especially during development.
```
com.auth0.auth.3z-j16G6OFMF_v-nt_QUZS5VV~vf.fGD:	{"nonce":"55yFu~kZUv0KT6azRtCIyf_c3dx-GAmJ"}
```
![image](https://cloud.githubusercontent.com/assets/8770194/24415185/ec2b9ea8-13e0-11e7-94d3-8b26c2d499c0.png)

Is there a reason why this could not be done via a single `localStorage` key, with JSON data?
Or possibly that in addition to `nonce`, a `issuedAt` field could be used for cleaning up old `state/nonce` combos?

luisrudge:
This is indeed inconvenient. I added this to our backlog and will tackle in the future.

csvn:
If it's something that you would appreciate a pull request on, you can mention it here. Not sure if there should be requirements about how long to store `state/nonce` items, if it should be configurable, etc.

luisrudge:
I'd love a PR. The thing is that I don't know what we should do in this case. What do you propose? add a `issuedAt` and clean the expired ones on start?

csvn:
Yeah, I think something like that might be fine for now.

1. Add a `issuedAt` field for the JSON object where `nonce` is stored
  ```
  "com.auth0.auth.<state>": "{ nonce: <nonce>, issuedAt: 1490783455 }"
  ```
2. (Possibly) Added a configurable field or default value to `auth0.js` options, for the number of hours/days before `state/nonce` pairs are removed
3. During `auth0.js` initialization, loop through localStorage keys for those prefixed with the namespace (`com.auth0.auth`), and remove outdated `state/nonce` pairs

csvn:
It would be possible to store all `state/nonce` pairs in a single namespace too, but it would require `JSON.parse` and `stringify` for each read/write. This due to the fact that `auth0.js` might be used from multiple browser tabs at the same time.

luisrudge:
Yeah. I like that. Let me talk with some people internally and I'll get back to you.

bragma:
Any news on this? Thanks!

[...]

PR #758, opened 2018-05-09
Clear local state when checkSession call fails

luisrudge:
partial fix for #402

### DIFF
--- a/src/web-auth/web-message-handler.js
+++ b/src/web-auth/web-message-handler.js
@@ -22,7 +22,8 @@ function runWebMessageFlow(authorizeUrl, options, callback) {
     timeoutCallback: function() {
       callback({
         error: 'timeout',
-        error_description: 'Timeout during executing web_message communication'
+        error_description: 'Timeout during executing web_message communication',
+        state: options.state
       });
     }
   });
@@ -58,10 +59,13 @@ WebMessageHandler.prototype.run = function(options, cb) {
   ) {
     var error = err;
     if (!err && eventData.event.data.response.error) {
-      error = objectHelper.pick(eventData.event.data.response, ['error', 'error_description']);
+      error = eventData.event.data.response;
+    }
+    if (!error) {
+      var parsedHash = eventData.event.data.response;
+      return _this.webAuth.validateAuthenticationResponse(options, parsedHash, cb);
     }
     if (
-      error &&
       error.error === 'consent_required' &&
       windowHelper.getWindow().location.hostname === 'localhost'
     ) {
@@ -69,11 +73,8 @@ WebMessageHandler.prototype.run = function(options, cb) {
         "Consent Required. Consent can't be skipped on localhost. Read more here: https://auth0.com/docs/api-auth/user-consent#skipping-consent-for-first-party-clients"
       );
     }
-    if (error) {
-      return cb(error);
-    }
-    var parsedHash = eventData.event.data.response;
-    _this.webAuth.validateAuthenticationResponse(options, parsedHash, cb);
+    _this.webAuth.transactionManager.clearTransaction(error.state);
+    return cb(objectHelper.pick(error, ['error', 'error_description']));
   });
 };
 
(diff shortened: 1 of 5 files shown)


src/web-auth/web-message-handler.js
```diff
@@ -58,22 +59,22 @@ WebMessageHandler.prototype.run = function(options, cb) {
...
       _this.warn.warning(
         "Consent Required. Consent can't be skipped on localhost. Read more here: https://auth0.com/docs/api-auth/user-consent#skipping-consent-for-first-party-clients"
       );
     }
-    if (error) {
-      return cb(error);
-    }
-    var parsedHash = eventData.event.data.response;
-    _this.webAuth.validateAuthenticationResponse(options, parsedHash, cb);
+    _this.webAuth.transactionManager.clearTransaction(error.state);
```
lbalmaceda: OK. This is the goal of this PR. However this `clearTransaction` won't get called if `error!=null` (see line 66, happy path). Is that fine?

src/web-auth/web-message-handler.js
```diff
@@ -58,22 +59,22 @@ WebMessageHandler.prototype.run = function(options, cb) {
   ) {
     var error = err;
```
lbalmaceda: this could be

```
var error = err ? err : eventData.event.data.response.error
if (!error) {
  parsedHash =..
  return validateAuthenticationResponse(...);
}

//continue handling errors
```

test/web-auth/web-auth.test.js
```diff
@@ -2402,7 +2404,7 @@ describe('auth0.WebAuth', function() {
       stub(IframeHandler.prototype, 'init', function() {
         this.timeoutCallback();
       });
-      this.auth0.checkSession({}, function(err, data) {
+      this.auth0.checkSession({ state: 'foobar' }, function(err, data) {
```
lbalmaceda: why has this changed? If I read the tests below (line 2452) it seems it's using this key as default state key value to remove when no error is provided (because of timeout). What would happen if this value is not given, like in the previous diff, when an error is also not provided (because of timeout)?

src/web-auth/web-message-handler.js
```diff
@@ -22,7 +22,8 @@ function runWebMessageFlow(authorizeUrl, options, callback) {
     timeoutCallback: function() {
       callback({
         error: 'timeout',
-        error_description: 'Timeout during executing web_message communication'
+        error_description: 'Timeout during executing web_message communication',
+        state: options.state
```
lbalmaceda: I find odd having an error object having "extra values" just for the sake of your code simplicity. It's easy to forget to filter the attributes returned to the user later.

src/web-auth/web-message-handler.js
luisrudge: Yes, because then we'll call `validateAuthenticationResponse`, which needs the transaction to work (and it will be removed later)

src/web-auth/web-message-handler.js
luisrudge: see `error = eventData.event.data.response;`the `eventData.event.data.response.error` property is a string.

src/web-auth/web-message-handler.js
luisrudge: the server also sends the same object `{error, error_description, state}`. I'm just replicating what the server does

test/web-auth/web-auth.test.js
luisrudge: Because the underlying code now depends on a state and my `stub` for the `transactionManager.process` method doesn't return a state: https://github.com/auth0/auth0.js/blob/6ac59c4faedc003820f3909aef5b5d57a572fad7/test/web-auth/web-auth.test.js#L2303

Since I already tested (elsewhere)[https://github.com/auth0/auth0.js/blob/6ac59c4faedc003820f3909aef5b5d57a572fad7/test/web-auth/web-auth.test.js#L2362] that we handle those params correctly, there's no need to stub the method again, so I just added the state

Issue #759, opened 2018-05-17
ReferenceError: global is not defined in Angular 6 application


fetis:
Hello, 

I'm facing this issue with fresh Angular 6 CLI project. The code to reproduce is very easy (config doesn't make sense here)
```typescript
    this.auth = new WebAuth({
      clientID: 'xxx',
      domain: 'xxx',
      responseType: 'token id_token',
      audience: 'xxx',
      redirectUri: `${location.protocol}//${location.host}/callback`,
      scope: 'openid'
    });

    this.auth.authorize();
```

and it produces
```
ERROR ReferenceError: global is not defined
    at Object.getWindow (window.js:12)
    at Object.randomString (random.js:9)
    at TransactionManager.push../node_modules/auth0-js/src/web-auth/transaction-manager.js.TransactionManager.generateTransaction (transaction-manager.js:45)
    at TransactionManager.push../node_modules/auth0-js/src/web-auth/transaction-manager.js.TransactionManager.process (transaction-manager.js:20)
    at WebAuth.push../node_modules/auth0-js/src/web-auth/index.js.WebAuth.authorize (index.js:581)
    at new AuthComponent (auth.component.ts:25)
    at createClass (core.js:10146)
    at createDirectiveInstance (core.js:10033)
    at createViewNodes (core.js:11255)
    at callViewAction (core.js:11571)
```
This's because of 
```js
function getWindow() {
  return global.window;
}
```
in window.js

According https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814
Angular team doesn't provide anymore any shimming for Node.js globals.

version: 9.5.1
Chrome 66 on MacOS, but I believe it's not browser specific.

fetis:
The quick and hacky workaround would be
```js
window.global = window
```
before any library usage

luisrudge:
duplicated https://github.com/auth0/auth0.js/issues/753#issuecomment-386774905

Issue #755, opened 2018-05-07
Shim for global has been removed from angular cli


sidneyleake:
I just started using your library and was following the tutorials. I decided to try to add auth0 to a brand new Angular application. While trying to get it to run I discovered that the latest Angular CLI removed the shim for global. This is used by the helper class in your library window.js. I just wanted to let you know in case you wanted to update your library to better interact with the latest version of angular.

luisrudge:
duplicated: https://github.com/auth0/auth0.js/issues/753

Issue #754, opened 2018-05-06
ReferenceError: global is not defined at Object.getWindow


ibenzyk:
Getting error:
when calling 

```
private _auth0 = new auth0.WebAuth({
    clientID: AUTH_CONFIG.CLIENT_ID,
    domain: AUTH_CONFIG.CLIENT_DOMAIN,
    responseType: 'token',
    redirectUri: AUTH_CONFIG.REDIRECT,
    audience: AUTH_CONFIG.AUDIENCE,
    scope: AUTH_CONFIG.SCOPE
  });

this._auth0.authorize();
```
I was following this guide [Auth guide](https://auth0.com/blog/real-world-angular-series-part-2/)

my `package.json`
```
{
  "name": "homemgmt",
  "version": "0.0.0",
  "license": "MIT",
  "scripts": {
    "ng": "ng",
    "start": "ng serve",
    "build": "ng build --prod",
    "test": "ng test",
    "lint": "ng lint",
    "e2e": "ng e2e"
  },
  "private": true,
  "dependencies": {
    "@angular/animations": "6.0.0",
    "@angular/cdk": "^6.0.0",
    "@angular/common": "6.0.0",
    "@angular/compiler": "6.0.0",
    "@angular/core": "6.0.0",
    "@angular/forms": "6.0.0",
    "@angular/http": "6.0.0",
    "@angular/material": "^6.0.0",
    "@angular/platform-browser": "6.0.0",
    "@angular/platform-browser-dynamic": "6.0.0",
    "@angular/router": "6.0.0",
    "@types/auth0-js": "^8.11.2",
    "auth0-js": "^9.5.1",
    "body-parser": "^1.18.2",
    "core-js": "^2.4.1",
    "cors": "^2.8.4",
    "express": "^4.16.3",
    "express-jwt": "^5.3.1",
    "jwks-rsa": "^1.2.1",
    "method-override": "^2.3.10",
    "mongoose": "^5.0.11",
    "rxjs": "^6.1.0",
    "zone.js": "^0.8.26"
  },
  "devDependencies": {
    "@angular/cli": "6.0.0",
    "@angular/compiler-cli": "6.0.0",
    "@angular/language-service": "6.0.0",
    "@types/jasmine": "~2.8.3",
    "@types/jasminewd2": "~2.0.2",
    "@types/node": "~6.0.60",
    "codelyzer": "^4.0.1",
    "jasmine-core": "~2.8.0",
[...]

Issue #753, opened 2018-05-04
global is not defined Angular 6


lucax88x:
Using any browser, by just performing a login with this library we get this error:

![image](https://user-images.githubusercontent.com/6294464/39647358-e7658aa6-4fde-11e8-940b-9211d6b9db27.png)

lucax88x:
as temporary workaround for any that reach this page you can add

` (window as any).global = window;`

in polyfill.ts

luisrudge:
Thanks for the workaround. At this point, the angular community will have to come with a solution for this problem, since we're hardly the only library using `global`. This was already discussed in #748. Moving away from using global means a huge refactor in auth0.js and it's not something we'll actively work on in the near future.

luisrudge:
More context on the angular change here: https://github.com/angular/angular-cli/issues/9827#issuecomment-369578814

luisrudge:
@fetis I think this is a bigger change than you think, but if you want to go for it, I'll review and test your PR 🎉

tinyweasel:
Workaround doesn't workaround for me.

fetis:
@tinyweasel you don't need workaround anymore. it's fixed since  v9.6.1

tinyweasel:
I just upgraded to Angular 7 and it's happening to me though. What should I update to fix it?

jayzhou3:
// app/app.js

```
import Application from '@ember/application';
import Resolver from './resolver';
import loadInitializers from 'ember-load-initializers';
import config from './config/environment';

window.global = window; //add this to your app.

const App = Application.extend({
  modulePrefix: config.modulePrefix,
  podModulePrefix: config.podModulePrefix,
  Resolver
});

loadInitializers(App, config.modulePrefix);

export default App;
```

ram4420:
i had same issue i just tried "npm install" it shows another error
"Error on Content Security Policy directive"
i was just added (window as any).global = window; in polyfilles.ts
now its working fine.

JobyCheriyan:
I too faced this issue after following an online training session, I resolved it by removing following element <router-outlet></router-outlet> from "app.component.html"

fetis:
@jayzhou3 @ram4420  the original issue originates to Angular and it was fixed. If you have an issue with other frameworks plz open another ticket and don't pollute this one.

jayzhou3:
[...]

PR #764, opened 2018-05-31
Remove global from window helpers

fetis:
Hello,

* I removed `global` from window helpers to make it Angular 6 compatible.
* Also, configured webpack to prevent issue `global` in the future (it does it automatically when encounters this variable).
* And some minor code cleanup.

I left current testing approach intact (manually mock `window` object). I think in the future it can be replaced with [JSDOM](https://github.com/jsdom/jsdom) or similar implementation

Related issues #759 #755 #754 #753 #748


luisrudge:
Thanks for the PR. Do you know why this line lost coverage? https://codecov.io/gh/auth0/auth0.js/compare/5a531f5013e77b0185e645954e5b505ab61b16db...2966805bd0100ede52a9aef1e81dd21ba5347d68/changes

fetis:
I assume it's because of this change https://github.com/auth0/auth0.js/pull/764/files#diff-fa3d40d414115b7a8cf9ef99f7282b22R173, which I took as someone accidentally duplicated the line.s

But I think it should be restored in a more explicit way, because popup_handler may return the current instance

luisrudge:
ahaha, yeah.. good catch! That's probably it, considering it's testing `should open the window once`

fetis:
I've fixed it. Show go back to normal now.

luisrudge:
Thanks for the PR! 🎉

fetis:
thanks for merge

luisrudge:
This was already released in 9.6.1 🎉

PR #772, opened 2018-06-08
Remove webpack and add Rollup as a module bundler

luisrudge:
Rollup is better suited to do module bundling for libraries. One of the reasons is because it allows us to easily target different types of bundles like UMD, ES and CJS. This PR only maintains the UMD bundle for now, but we can add ES modules later. The benefit of using ES modules later is that consumers of this library will be able to do "tree shaking", which means their bundle will only include what they use, everything else will be left off. This can greatly reduce the bundle size for people only using `WebAuth` and not `Authentication`, for example.

Here's a great article about the difference between webpack and rollup for libraries:
https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c

Summary of changes:
- Files will go to both the `dist` and `build` folder
- Changed CDN to read files from the `dist` folder instead of `build`
- Remove all gulp and webpack related packages
- added `browser` field in package.json which specifies alternative files to load for people bundling for the browser
- added `files` field in package.json so we deploy the distribution files to the npm registry (it will allow users to use services like https://unpkg.com/)

## Before

image

## After

image


joshcanhelp:
@luisrudge - Thank you for the detailed run-down! Informative screenshots and good supporting article:

>If you need code-splitting, or you have lots of static assets, or you’re building something with lots of CommonJS dependencies, Webpack is a better choice. If your codebase is ES2015 modules and you’re making something to be used by other people, you probably want Rollup.

I'm definitely sold that this is the right way to go and I don't know how helpful an in-depth config file review would be. My main concern here, though, is what we talked about last week: the change from `build` to `dist`. I agree that `dist` is generally the "standard" directory, and the one I typically use myself, but is this change required for Rollup? Is it considered breaking for anyone? Should it be done in a separate PR? 

In short: is there a case where changing this output folder will cause a break in someone's deployment procedure, even if it's in the minority of cases?

rollup.config.js
```diff
@@ -0,0 +1,99 @@
...
+
+const getPlugins = isProduction => {
+  return [
+    resolve({
+      browser: true
+    }),
+    commonjs(),
+    json(),
+    replace({
+      __DEV__: isProduction ? 'false' : 'true',
```
joshcanhelp: `! isProduction `

rollup.config.js
```diff
@@ -0,0 +1,99 @@
...
+      sourcemap: true,
+      exports: 'named'
+    },
+    plugins: getPlugins(isProduction)
+  },
+  {
+    input: 'plugins/cordova/index.js',
+    output: {
+      name: 'CordovaAuth0Plugin',
+      file: 'dist/cordova-auth0-plugin.min.js',
```
joshcanhelp: Might be better to have `dist/` as a var

luisrudge:
> I'm definitely sold that this is the right way to go and I don't know how helpful an in-depth config file review would be. My main concern here, though, is what we talked about last week: the change from build to dist. I agree that dist is generally the "standard" directory, and the one I typically use myself, but is this change required for Rollup? Is it considered breaking for anyone? Should it be done in a separate PR?
> In short: is there a case where changing this output folder will cause a break in someone's deployment procedure, even if it's in the minority of cases?

Yeah, I'll change to use build and dist.

> Edit: one more that came up from the article ... is there any reason to include a pkg.module declaration in package.json?
Once we have ES modules fully supported, then yes

rollup.config.js
```diff
@@ -0,0 +1,99 @@
...
+    plugins: getPlugins(isProduction)
+  },
+  //dev
+  {
+    input: 'src/index.js',
+    output: {
+      name: 'auth0',
+      file: 'dist/auth0.js',
+      format: 'umd',
+      sourcemap: isProduction ? false : 'inline',
```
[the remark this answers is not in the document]
luisrudge: `isProduction` is more like: `areWeBuildingAProductionFile`

rollup.config.js
luisrudge: Not sure what that means 😬

rollup.config.js
luisrudge: probably 😬

rollup.config.js
joshcanhelp: Simpler statement, same meaning

luisrudge: nope, I need the result to be a string, so it'd have to be either `(!isProduction).toString()` or what I have there

PR #774, opened 2018-06-09
Migrate to ES modules

luisrudge:
In order to be able to properly do tree-shaking, we need to start using ES modules (`import` / `export`) instead of CommonJS modules (`require`). This PR is a find/replace in all files, making them use import/export. This, of course, broke all the tests 😬. `mocha`, our test runner, doesn't support modern javascript, so I needed to add `babel` to transpile the code to compatible javascript during testing.

### Summary of changes:

- Added `babel` to transpile test code in order to make `mocha` work
- Upgraded eslint auth0 config so it allows modern javascript. Since we're not migrating our code yet, I had to turn off [a bunch of rules](https://github.com/auth0/auth0.js/pull/774/files#diff-1dc6ee56b778cd91e0327b52aaeaa1b9R19)
- change istanbul (our test coverage runner) to nyc. nyc is pretty much istanbul 2.0 (see [here](https://github.com/istanbuljs/nyc/issues/524#issuecomment-280979372))
- upgraded prettier, so we might have a few automatic styling changes
- upgrade circle-ci to node@v8.9.0 because of a dependency


.babelrc
```diff
@@ -0,0 +1,8 @@
+{
```
luisrudge: because this is not supported by mocha, we use babel to convert the code to something that mocha understands

.eslintrc
```diff
@@ -16,6 +15,19 @@
     "no-param-reassign": "off",
     "consistent-return": "off",
     "vars-on-top": "off",
-    "compat/compat": "error"
+    "compat/compat": "error",
```
luisrudge: because import/export (es6 features) weren't supported by our super old eslint rules, I had to update them and disable everything that was against those rules.

.nycrc
```diff
@@ -0,0 +1,8 @@
+{
```
luisrudge: istanbul is now nyc, so I updated the package

package.json
```diff
@@ -26,10 +26,10 @@
   "scripts": {
     "start": "rollup -c --watch",
     "build": "rm -rf dist && rm -rf build && rollup -c --prod && cp -rf dist build",
-    "test": "mocha test/**/*.test.js test/setup.js",
-    "test:integration": "mocha-parallel-tests --max-parallel 2 integration/**/*.test.js",
-    "test:watch": "mocha --watch -R min test/**/*.test.js  test/setup.js",
-    "ci:test": "istanbul cover _mocha --report lcovonly -R test/**/* test/setup.js -- -R mocha-multi --reporter-options spec=-,mocha-junit-reporter=-",
+    "test": "cross-env BABEL_ENV=test mocha --require babel-register test/**/*.test.js --exit",
```
luisrudge: cross-env makes sure we can run tests in both windows and unix

luisrudge:
requires in those folders are ok to have. I didn't change the way the integration tests run, so it's fine.

 Can't do the beautify in other pr because it's already automated in precommit hook.

joshcanhelp:
> Can't do the beautify in other pr because it's already automated in precommit hook.

It's beautifying unchanged code, though. You mentioned that this was because you upgraded the package. Did that upgrade have to happen in this changeset?

luisrudge:
OH yeah, no. I could've updated this in another branch.. but I had to update the eslint config and one thing led to the other.. sorry 😬

edmorley:
Hi!

> In order to be able to properly do tree-shaking

You may be aware of this already, but just in case - for `target: 'web'` webpack's [mainFields](https://webpack.js.org/configuration/resolve/#resolve-mainfields) setting defaults to `['browser', 'module', 'main']`. This means that the current auth0-js package.json `browser` value of `"dist/auth0.min.js"` will take priority over the `main` value of `"src/index.js"`, and so no tree shaking will take place.

It seems like the correct package.json settings would be:

```js
  "main": "dist/auth0.min.js",
  "module": "src/index.js",
```
(with `"browser"` not defined at all)

...which would then allow tree shaking to work.

luisrudge:
@edmorley this pr was so we can tree shake the code we use (as in, our dependencies), not to make auth0js “tree shakable”. But I wasn’t aware of the “browser” thing though. Do you have any articles on that? I want to be sure that tree shake works with our lib as well.

edmorley:
Ah makes sense. The unminified `build/auth0.js` did decrease slightly between v9.6.1 and v9.7.2 (314KB -> 299KB), although the minified file (`build/auth0.min.js`) did actually increase in size (112KB -> 138KB) - possibly due to different uglify options.

Re tree-shaking this package itself, I'd check out:
https://rollupjs.org/guide/en#publishing-es6-modules
https://webpack.js.org/guides/tree-shaking/

At the moment, since webpack will use `build/auth0.min.js` (which uses CJS modules) any consumers of the package won't be able to tree shake the parts of it they are not using.

luisrudge:
I did some size comparisons when we migrated to rollup and we saw a size decrease in all the files: https://github.com/auth0/auth0.js/pull/772

I’ll read those links next week. Thanks!

PR #783, opened 2018-06-27
Don't open a popup when doing sign ups

fbarbare:
These changes fix an issue where when using the `popup.signupAndLogin` it is opening a Popup for no reason. That popup is not being used at all on sign-up phase.


luisrudge:
This makes sense. We changed the way the login happens (no popup anymore). Can you please write a test for this?

fbarbare:
Hi @luisrudge , I was thinking about it but does it make sense to test something that does not happen at all? Because if we start testing those, we could be here a while 😂

luisrudge:
@fbarbare I agree, yeah. Can you write a specific test for this line we're missing? https://codecov.io/gh/auth0/auth0.js/compare/87cb2d3f530e82f56636e1b5afba815e4d7de3a2...e33f90e745131fd7c3a17e45c9d3c049ca7f2cb8/changes#D1-70

luisrudge:
ping @fbarbare

fbarbare:
I've added the tests to the PR.
Thanks @MatthewBacalakis

luisrudge:
Thanks for all the effort 🎉

PR #784, opened 2018-06-28
Handle remove child error

claudiahdz:
Added a tiny verification for the parent node since some customers where seeing errors in which the node didn't exist.

ex.
The following line of code emit multiple errors at different situations:
`_window.document.body.removeChild(_this.iframe);`

The errors include:
Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.
The object can not be found here.
Node was not found

The error message is:
`Cannot read property 'body' of undefined``

ZENDESK Ticket: https://auth0.zendesk.com/agent/tickets/43507


luisrudge:
Can you please write a test for this? Thanks!

Issue #740, opened 2018-04-13
Refactor how we process transactions


luisrudge:
```js
var params = objectHelper
    .merge(this.baseOptions, [
      'clientID',
      'responseType',
      'redirectUri',
      'scope',
      'audience',
      '_csrf',
      'state',
      '_intstate',
      'nonce'
    ])
    .with(options);
  params = this.transactionManager.process(params);
```

snippets like this are all over the place. Ideally, we should abstract this so we don't have to copy/paste everywhere

luisrudge:
test

lbalmaceda:
Worth noting that as of today, if you need to add or remove a value from that array you'll need to change it everywhere.

luisrudge:
With #787, it was pretty clear this is a deeper issue with how we pick and choose which param goes in and out. Needs more thought and it will probably break users, so we'll leave this to the next major as a feature request.

PR #787, opened 2018-07-02
Proposal to refactor how transactions are processed

jessyao26:
In response to issue #740 
**Implementation**
I went through /web-auth/index.js and saw duplicate param arrays `[
  'redirectUri',
  'responseType',
  'scope',
  'audience',
  '_csrf',
  'state',
  '_intstate',
  'nonce'
]` 
Some param arrays were passing in slightly different strings for different prototype methods (i.e. responseMode). I created a 'base' param array that gets concated with custom values at the top of the index.js file.
**More ideal implementation**
Ideally, I would create a file of constants separate from this index.js file and require it whereever needed. This file of constants could help fill in vars with vscode's intellisense and eliminate magic strings, but I didn't want to make any large changes.
**Notes**
-There's two unrelated lines that have different spacing. I didn't touch these lines. I'm unsure if it's a prettifier or minifier that's adding spacing.
-I tried to use the 'const' keyword instead of vars, when setting arrays, but I think I'm pointing to an older version of es6
-There's another similar params array inside  /web-auth/popup.js, but it has different params, if there was a constants file (from the more ideal implementation section) this would be a use case
**Feedback**
let me know if there's anything that can be improved! Or if you're passing on this, feel free to let me know why!

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,11 +52,11 @@
     "winchan": "^0.2.0"
   },
   "devDependencies": {
+    "@auth0/component-cdn-uploader": "auth0/component-cdn-uploader#v2.1.1",
     "babel-plugin-istanbul": "^4.1.6",
     "babel-preset-env": "^1.7.0",
     "babel-register": "^6.26.0",
     "codecov": "^3.0.2",
-    "component-cdn-uploader": "auth0/component-cdn-uploader#v2.1.0",
     "cross-env": "^5.1.6",
     "eslint": "^4.19.1",
     "eslint-config-auth0-base": "^13.1.0",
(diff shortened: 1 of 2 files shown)


luisrudge:
Thanks @jessyao26 for your PR! I think we should go with your ideal implementation here, because we actually use this kind of baseOptions merge in a [a lot](https://github.com/auth0/auth0.js/search?l=JavaScript&q=objectHelper.merge%28this.baseOptions) of places, so abstracting away this specific code makes a lot of sense.

jessyao26:
see #792 !

PR #789, opened 2018-07-06
options is optional in WebAuth.prototype.authorize

behrangsa:
In your tutorials, `authorize` is called without any arguments ([example](https://auth0.com/docs/quickstart/spa/jquery/01-login#create-an-authentication-service)).

Consequently, `options` should be marked as optional in the JsDoc (i.e. `[options]`), otherwise IDEs might warn that its invocations require 1 argument:

screenshot from 2018-07-06 23-03-25


behrangsa:
Looks like there are quite a few functions like this in the code base:

* `WebAuth.prototype.parseHash = function(options, cb) {`
* `WebAuth.prototype.renewAuth = function(options, cb) {`

Might be better to reject this PR and fix all of the JsDocs in a single commit.

behrangsa:
@luisrudge I updated the PR with the other occurrences I could find.

luisrudge:
Thanks!

PR #792, opened 2018-07-07
Refactor params

jessyao26:
**Implementation**
I added a constants files separated into categories of different constants (i.e. oauth2, auth0, blacklist).
In another file, I created a params object that contains param arrays with separated by property for each type of param array. 
**Note**
The params object makes more sense when it's not too nested like paramsArray.blacklist.popup.handler.base. It gets really nested with certain properties because their param arrays only differ by one or two values. I'm worried this is confusing, but am unsure of a different way to implement it right now. Let me know what you think, feel free to give feedback


luisrudge:
I agree this is confusing (when you use a lot of them), we should probably just use the same values all around, but it's hard to take a decision on which params to use right now. I'll evaluate this approach during this week (I'm super busy this week, sorry in advance).

luisrudge:
I'm sorry it took that much time to review this. After talking with some colleagues, we decided that this approach would be very confusing with the current state of the SDK. The real issue is that we have a lot of different overrides all around and that is the main source of confusion. Thanks so much for your effort and I'm sorry we're not going through with this amazing PR. 🎉

Issue #806, opened 2018-07-16
Upgrading to 9.7.2 causes test breakages


murtyjones:
I upgraded from auth0-js 9.5.1 to ^9.7.2 and am now seeing the following error when running my jest tests:

```
/Users/[my-user]/[my-repo]/node_modules/auth0-js/src/index.js:1
({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import Authentication from './authentication'
```

luisrudge:
What's the error? I think your string was trimmed.

murtyjones:
My bad:

```
/Users/.../.../node_modules/auth0-js/src/index.js:1
({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,global,jest){import Authentication from './authentication';
                                                                                             ^^^^^^

SyntaxError: Unexpected token import
```

luisrudge:
From what I can see, it appears that your code is:
- not using our bundled version (which is referenced here: https://github.com/auth0/auth0.js/blob/master/package.json#L8)
- not transpiling ES Modules (import/export) to commonjs modules (require/module.exports)

Either one of the two will fix this

murtyjones:
OK. I added this to my jest config:
```
"transformIgnorePatterns": [
      "/node_modules/(?!auth0-js).+\\.js$"
    ]
```

Which fixed it. 

Out curiosity, how would I use the bundled version with React?

Timer:
9.7.0 should've been a major version bump, as any application participating in SSR is now broken.
I just got bit by this in a Next.js application.

luisrudge:
@Timer It wasn’t our intention to break apps and I want to fix this in a patch release if we did that. I’ll recreate this scenario tomorrow to make sure it still works. My understanding of this is that apps would get the “browser“ npm entry and would all work out. But, as you pointed out, maybe not all apps are doing that. Just so you can help me testing that: does this issue happen on a brand new next js app + auth0js?

Timer:
Great to hear, @lucascrespo! Looking forward to a fix. 😄 

Apps will look at the `main` field when running via Node (SSR, or tests -- like this issue) and the `browser` or `module` field when being consumed by webpack (normal bundle application).

General rule of thumb is as follows:
  
  - `esnext`: uncompiled sources, what `main` is currently set to
  - `browser`: compiled down to minimum supported browser version (and no ES Modules)
[...]

PR #808, opened 2018-07-18
Fix npm module export

luisrudge:
- Remove rollup-plugin-uglify because it doesn't support minifying ES6+. Using rollup-plugin-terser instead
- Don't generate prod files when in dev mode
- instead of using main/browser in package.json, using module/main (ES modules, UMD bundle)
	- This follows best practices outlined [here](https://github.com/auth0/auth0.js/issues/806#issuecomment-405430536), [here](http://2ality.com/2017/04/setting-up-multi-platform-packages.html#solutions) and [here](https://github.com/rollup/rollup-starter-lib/blob/master/package.json#L5-L6).

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,6 @@
     "rollup-plugin-serve": "^0.4.2",
     "rollup-plugin-sourcemaps": "^0.4.2",
     "rollup-plugin-terser": "^1.0.1",
-    "rollup-plugin-uglify": "^4.0.0",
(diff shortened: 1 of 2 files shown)


rafaelrinaldi:
FYI `9.7.2` was totally broken for me and this PR fixed it. I'm using the beta version for now.

luisrudge:
@rafaelrinaldi Thanks for letting us know. Can you tell me your scenario so I can compare with others?

Issue #816, opened 2018-08-01
co/verifier localStorage key is not being removed after a successful login


luisrudge:

thierryMic:
@luisrudge i am using your [nextJs example](https://github.com/luisrudge/next.js-auth0) i added
`localStorage.removeItem(`com.auth0.auth.${secret}`)` to the componentDidMount function of the call-back (signed-in page).

It works well for removing the nonce after successful log-ins but nonces keep building up for log-outs.

Do you agree with the approach for removing nonces on succesful log-ins and do you have any recommendations to get rid of the nonces that are created on logout?

luisrudge:
This has to be fixed inside our library. You can add something to remove now but, ultimately, it's a bug we'll fix.

luisrudge:
This was fixed by switching to a cookie with a short expiration date #817

PR #817, opened 2018-08-07
Start using cookies instead of localStorage by default

### LINKED ISSUE #402: localStorage is spammed with state/nonce combos

luisrudge:
fix #402 #816 

In some cases, there's no way to old local storage entries because we don't have enough information:

- if the auth result wasn't successful
- if the cross origin auth was successful, but it didn't use the cross auth callback page

Because of that, cookies are a better fit for storage, since it has built-in expiration dates.

There's a new option `__tryLocalStorageFirst` that restores the previous behavior if you find any compatibility issues (please report them if you find it) or prefer to use localStorage for some reason.

This is not considered a breaking change because there shouldn't be any behavioral changes between versions.

### DIFF
--- a/src/helper/ssodata.js
+++ b/src/helper/ssodata.js
@@ -1,18 +1,22 @@
-import storage from './storage';
+import Storage from './storage';
 
-export default {
-  set: function(connection, sub) {
-    var ssodata = {
-      lastUsedConnection: connection,
-      lastUsedSub: sub
-    };
-    storage.setItem('auth0.ssodata', JSON.stringify(ssodata));
-  },
-  get: function() {
-    var ssodata = storage.getItem('auth0.ssodata');
-    if (!ssodata) {
-      return;
-    }
-    return JSON.parse(ssodata);
-  }
+function SSODataStorage(options) {
+  this.storage = new Storage(options);
+}
+
+SSODataStorage.prototype.set = function(connection, sub) {
+  var ssodata = {
+    lastUsedConnection: connection,
+    lastUsedSub: sub
+  };
+  this.storage.setItem('auth0.ssodata', JSON.stringify(ssodata));
 };
+SSODataStorage.prototype.get = function() {
+  var ssodata = this.storage.getItem('auth0.ssodata');
+  if (!ssodata) {
+    return;
+  }
+  return JSON.parse(ssodata);
+};
+
+export default SSODataStorage;
--- a/src/helper/storage/handler.js
+++ b/src/helper/storage/handler.js
@@ -6,7 +6,7 @@ import Warn from '../warn';
 function StorageHandler(options) {
   this.warn = new Warn({});
   this.storage = new CookieStorage();
-  if (!options.__tryLocalStorageFirst) {
+  if (options.__tryLocalStorageFirst !== true) {
     return;
   }
   try {
--- a/src/helper/storage/handler.js
+++ b/src/helper/storage/handler.js
@@ -3,9 +3,12 @@ import DummyStorage from './dummy';
 import CookieStorage from './cookie';
 import Warn from '../warn';
 
-function StorageHandler() {
+function StorageHandler(options) {
   this.warn = new Warn({});
   this.storage = new CookieStorage();
+  if (!options.__tryLocalStorageFirst) {
+    return;
+  }
   try {
     // some browsers throw an error when trying to access localStorage
     // when localStorage is disabled.
--- a/test/helper/storage-handler.test.js
+++ b/test/helper/storage-handler.test.js
@@ -28,8 +28,12 @@ describe('helpers storage handler', function() {
   afterEach(function() {
     windowHandler.getWindow.restore();
   });
-  it('should use localStorage by default', function() {
-    var handler = new StorageHandler();
+  it('should use CookieStorage by default', function() {
+    var handler = new StorageHandler({});
+    expect(handler.storage).to.be.a(CookieStorage);
+  });
+  it('should use localStorage when __tryLocalStorageFirst is true', function() {
+    var handler = new StorageHandler({ __tryLocalStorageFirst: true });
     expect(handler.storage).to.be.a(MockLocalStorage);
   });
   describe('when using localStorage', function() {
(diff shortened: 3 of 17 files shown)


src/helper/storage/handler.js
```diff
@@ -3,9 +3,12 @@ import DummyStorage from './dummy';
 import CookieStorage from './cookie';
 import Warn from '../warn';
 
-function StorageHandler() {
+function StorageHandler(options) {
   this.warn = new Warn({});
   this.storage = new CookieStorage();
+  if (!options.__tryLocalStorageFirst) {
```
joshcanhelp: Do we want to be explicit about what `options.__tryLocalStorageFirst` can be? This is truth-y, should it be `if (!options.__tryLocalStorageFirst === true)` instead? Edge case and a new option so no existing behavior but `"no"` or `"false"` would trigger a positive case here. Just wondering how this check is done elsewhere.

test/helper/storage-handler.test.js
```diff
@@ -55,30 +59,30 @@ describe('helpers storage handler', function() {
...
       var key = 'foo';
-      var handler = new StorageHandler();
+      var handler = new StorageHandler({ __tryLocalStorageFirst: true });
       handler.removeItem(key);
       expect(removeItemSpy.firstCall.args).to.be.eql([key]);
     });
   });
   describe('should use cookie storage', function() {
-    it('when localstorage is not available', function() {
+    it('when __tryLocalStorageFirst is true but localSTorage is not available', function() {
```
joshcanhelp: "localStorage"

src/helper/ssodata.js
```diff
@@ -1,18 +1,22 @@
...
-    if (!ssodata) {
-      return;
-    }
-    return JSON.parse(ssodata);
+function SSODataStorage(options) {
+  this.storage = new Storage(options);
+}
+
+SSODataStorage.prototype.set = function(connection, sub) {
+  var ssodata = {
```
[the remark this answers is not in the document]
luisrudge: auth0.js is not transpiling code yet, so no let or const are allowed.

test/helper/storage-handler.test.js
luisrudge: CAN I USE ALL CAPS LIKE THIS LOCALSTORAGE WHAT DO YOU THINK

src/helper/storage/handler.js
luisrudge: done

test/helper/storage-handler.test.js
luisrudge: fixed 😬

joshcanhelp: Either all caps or camel, pick one! 😆

mythmon:
> This is not considered a breaking change because there shouldn't be any behavioral changes between versions.

I have a question about this. Before this change, auth details were stored in localstorage, and kept entirely on the client. After this change, the auth details are stored in cookies. If I understand correctly, all of the cookies on the client are sent to the server for each request. This bloats the request size, and could represent a credentials leak.

Is my understanding above wrong? If not, this does not seem a change without behavioral consequences. I was extremely surprised to see it in a minor release of the library.

luisrudge:
There are no auth details stored in cookies. We only store:

One cookie for (expires in 24 hours):
- last used connection
- last user id that logged in

One cookie for (expires in 30 minutes):
- nonce (auto generated or provided by the developer)
- state (auto generated or provided by the developer)

One cookie for (expires in 15 minutes):
- cross origin verification key

Yes, the cookies increase the overall request size and that's why we made the transactional cookies short lived. We considered that slightly increasing the payload size was a better trade off than possibly making our customers' apps crash due to localStorage being full.

sholladay:
FYI, cookies are _more_ prone to size and storage quantity limitations.
 - http://browsercookielimits.squawky.net/
 - https://stackoverflow.com/questions/2989284/what-is-the-max-size-of-localstorage-values

One notable difference is that if cookie storage is full, it won't cause any JavaScript errors (to my knowledge), whereas `locatStorage.setItem()` will throw. I would probably prefer the `localStorage` exception with a known error type (which can be ignored explicitly) over a silent failure. Maybe there are other good reasons to use cookies, though, like the expiration system.

src/helper/storage/handler.js
```diff
@@ -3,9 +3,12 @@ import DummyStorage from './dummy';
 import CookieStorage from './cookie';
 import Warn from '../warn';
 
-function StorageHandler() {
+function StorageHandler(options) {
   this.warn = new Warn({});
   this.storage = new CookieStorage();
+  if (options.__tryLocalStorageFirst !== true) {
```
jbeurel: Hello @luisrudge. How can I use LocalStorage instead of CookieStorage in a TypeScript application? It seems impossible to set __tryLocalStorageFirst to `true` because of `AuthorizeOptions` or `AuthOptions` type definition: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/auth0-js/index.d.ts#L166

luisrudge: send a PR to the typing repo adding this flag?

luisrudge: Also, why you want to use localStorage instead of cookies? that would be pretty helpful to understand, since we're considering removing localStorage in the future.

jbeurel: Hello @luisrudge

"send a PR to the typing repo adding this flag?", Yes. PR done and merged by DefinitelyTyped team 👍

"Also, why you want to use localStorage instead of cookies?" Because I did not find how to make auth0 cookie work with Electron.

luuuis:
One customer was concerned about these cookies.

> Customer [...] has noticed a cookie appearing on their domain, com.auth0.auth.testNonce3 - what is this used for? Is it a test that has leaked onto production somehow?

https://auth0team.atlassian.net/projects/ESD/queues/custom/405/ESD-1523

PR #831, opened 2018-10-03
Update .gitignore

darshkpatel:
Added VSCode settings directory ```.vscode``` which isn't required by the project.


luisrudge:
It's not required, but we leave it in the repo to make it easier for contributors using that editor.

PR #835, opened 2018-10-05
fix(TransactionManager): Set state to expire in 30 mins

sayuti-daniel:
Passed `expires` option does not override the default value.
I'd like to ask is it okay to store nonce for 30 mins? IMO, a very short time around 5 mins is enough.

### DIFF
--- a/src/web-auth/transaction-manager.js
+++ b/src/web-auth/transaction-manager.js
@@ -54,7 +54,7 @@ TransactionManager.prototype.generateTransaction = function(
       state: state,
       lastUsedConnection: lastUsedConnection
     },
-    times.MINUTES_30
+    { expires: times.MINUTES_30 }
   );
   return {
     state: state,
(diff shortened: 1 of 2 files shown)


sayuti-daniel:
This has been introduced since `9.5.0` :smile: I'm curious to know why the cookie state is not removed instead of using expiration time. Do you mind explaining this? Thanks! :beers:

luisrudge:
Thanks so much for the PR 🎉 

We remove the transaction state once the authentication is successful. But we can't remove if the authentication fails for any reason, because we don't get a `state` parameter back in that case (the `state` param is used to "find" the transaction state in the cookie).

sayuti-daniel:
@luisrudge Can we have patch release for this?

luisrudge:
@sayuti-daniel we'll have a new release in 2 days :shipit:

PR #846, opened 2018-10-23
Parse file protocol from Url

anion155:
When this library is used in electron, locations in most cases are files


luisrudge:
Can you please add a test for that? Thanks!

luisrudge:
ping @anion155

anion155:
Will add. In a moment.

anion155:
There is 'url' helper, used only once in 'web-auth/popup', shouldn't it be 'objectHelper.getOriginFromUrl'?

PR #854, opened 2018-10-25
Create PULL_REQUEST_TEMPLATE.md

sakethramanujam:
I guess there is a need for one such PR Templates. It, makes the life of maintainers easy.


luisrudge:
Thanks for the PR, but we're actually in the middle of adding templates to our repos. See here for an example: https://github.com/auth0/lock/pull/1524

PR #859, opened 2018-10-26
Better gitignore file & Better issue and pull request templates

filips123:
Add better gitignore file.
Add better  issue and pull request templates.


luisrudge:
Thanks, but we just updated our templates.

filips123:
@luisrudge What about updating only gitignore file?

luisrudge:
Current file is fine for now, thanks

PR #864, opened 2018-10-27
Refactoring to ES6

Archaeologist03:


luisrudge:
Thanks, but not something we’re interes at the moment

PR #867, opened 2018-10-30
Added semicolons to jsdoc-default.css

astv99:
Added semicolons to /docs/styles/jsdoc-default.css on various lines where semicolons were not already in place, to ensure that all CSS rules were properly formed.


luisrudge:
This file is auto generated

PR #891, opened 2019-01-12
Add `test:coverage` npm script to check coverage locally

ScottRudiger:
Currently, there is no standard way to check coverage locally.

From the README:
> Run npm run test:coverage to run the test suite with coverage report.

This PR simply adds a `test:coverage` npm script that checks coverage via nyc.

Happy to change up the script if there's a different variation you'd like to run. Alternatively, we could remove the above line from the README instead of adding a script.


ScottRudiger:
I just noticed there's also the `ci:test` script, which does *almost* the same thing.

Should we add the `test:coverage` script anyway to remain consistent with the README and as a distinction between development & CI?

Issue #872, opened 2018-11-08
Request to Auth0 should trim username/email


luisrudge:

luisrudge:
closed in https://github.com/auth0/auth0.js/pull/895

PR #895, opened 2019-01-18
Trim `username`, `email` and `phoneNumber` params in every request

ScottRudiger:
This is in reference to issue #872

It's possible there are additional places that username/email should be trimmed. If so, you can consider this a WIP 😉; but as I'm still getting used to the code base, I wanted to get this in to make sure I'm on the right track.


src/helper/request-builder.js
```diff
@@ -45,7 +46,7 @@ RequestObj.prototype.set = function(key, value) {
 };
 
 RequestObj.prototype.send = function(body) {
-  this.request = this.request.send(body);
+  this.request = this.request.send(objectHelper.trimUsernameAndEmail(body));
```
luisrudge: please add a test for this

src/helper/object.js
```diff
@@ -147,6 +147,22 @@ function getOriginFromUrl(url) {
...
+  }
+  return trimmed;
+}
+
+function trimMultiple(options, keys) {
+  return keys.reduce(trim, options);
+}
+
+function trimUsernameAndEmail(options) {
+  return trimMultiple(options, ['username', 'email']);
```
luisrudge: we should probably trim `phoneNumber` as well

luisrudge:
Thanks @ScottRudiger! that's amazing! I left a couple of comments of things that I think are missing.

luisrudge:
Thanks, that's perfect 🎉

PR #910, opened 2019-02-21
coerce undefined to null

mattkrick:
impersonation is broken for me because of a state mismatch, i traced it to `undefined === null`.


luisrudge:
Thanks, do you know why the test is failing? Can you write a new test to make sure we're handling both cases?

luisrudge:
Hi! I'm sorry about the copy/paste message, but I'm cleaning up some stale issues / PRs. 🗑 
It's been a while since this PR was opened, but we didn't see any traction after our last interaction. If you have more information to add, feel free to answer this PR. Thanks! 🎉

Issue #836, opened 2018-10-10
webAuth.popup.authorize is inconsistent with webAuth.authorize


Pyrolistical:
auth0.js version 9.8.0

When using `webAuth.popup.authorize`, the callback will have an `authResult.idTokenPayload` with the user profile. The profile will contain `appMetadata`.

When using `webAuth.authorize`, then using `webAuth.parseHash` upon the redirect, the callback will have an `authResult.idTokenPayload` with the user profile. The profile will contain `app_metadata`.

This is very confusing by switching between redirect and popup causes the profile to be camel cased or not. This does not only apply to app_metadata, it applies to all fields on the profile.

This is especially annoying when using email/password database as that must be redirect only and if we use social login as pop, we need handle both form of casing.

mariosant:
The same applies for `webAuth.checkSession` as well. This is because of using the following as a convention:
https://github.com/auth0/auth0.js/blob/master/src/helper/response-handler.js#L79

This feels like an early design issue - people were not sure how they should serve/handle stuff, which lead to such inconsistencies. I am curious, how should this one be fixed?

luisrudge:
This will be a hard one to solve without a breaking change 🤔

I'll add this to my near term backlog so I can take a deep look.

PR #911, opened 2019-02-25
Fix different id_token payload casing between authorize and popup.authorize

luisrudge:
fix https://github.com/auth0/auth0.js/issues/836

### Changes

- Added new option to the `toCamelCase` helper function to keep the original values when converting to camel case
- Used this new option when calling `popup.authorize`

### References

Please include relevant links supporting this change such as a:

https://github.com/auth0/auth0.js/issues/836

### Testing

[x] This change adds unit test coverage

[ ] This change adds integration test coverage

### Checklist

[x] I have read the [Auth0 general contribution guidelines](https://github.com/auth0/open-source-template/blob/master/GENERAL-CONTRIBUTING.md)

[x] I have read the [Auth0 Code of Conduct](https://github.com/auth0/open-source-template/blob/master/CODE-OF-CONDUCT.md)

[x] All tests and linters described in the [Develop section](https://github.com/auth0/auth0.js#develop) run without errors

### DIFF
--- a/src/helper/object.js
+++ b/src/helper/object.js
@@ -103,16 +103,21 @@ function toSnakeCase(object, exceptions) {
   }, {});
 }
 
-function toCamelCase(object, exceptions) {
+function toCamelCase(object, exceptions, options) {
   if (typeof object !== 'object' || assert.isArray(object) || object === null) {
     return object;
   }
 
   exceptions = exceptions || [];
-
+  options = options || {};
   return Object.keys(object).reduce(function(p, key) {
     var newKey = exceptions.indexOf(key) === -1 ? snakeToCamel(key) : key;
-    p[newKey] = toCamelCase(object[key]);
+
+    p[newKey] = toCamelCase(object[newKey] || object[key], [], options);
+
+    if (options.keepOriginal) {
+      p[key] = toCamelCase(object[key], [], options);
+    }
     return p;
   }, {});
 }
--- a/src/helper/response-handler.js
+++ b/src/helper/response-handler.js
@@ -76,7 +76,10 @@ function wrapCallback(cb, options) {
       return cb(null, data.body || data);
     }
 
-    return cb(null, objectHelper.toCamelCase(data.body || data));
+    return cb(
+      null,
+      objectHelper.toCamelCase(data.body || data, [], { keepOriginal: options.keepOriginalCasing })
+    );
   };
 }
 
--- a/example/callback_popup.html
+++ b/example/callback_popup.html
@@ -1,20 +1,18 @@
 <!DOCTYPE html>
 <html>
+  <head>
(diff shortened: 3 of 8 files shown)


src/helper/object.js
```diff
@@ -103,16 +103,21 @@ function toSnakeCase(object, exceptions) {
   }, {});
 }
 
-function toCamelCase(object, exceptions) {
+function toCamelCase(object, exceptions, options) {
```
joshcanhelp: Do you really need all the options passed in? Or just boolean `keepCasing`?

src/helper/response-handler.js
```diff
@@ -76,7 +76,10 @@ function wrapCallback(cb, options) {
       return cb(null, data.body || data);
     }
 
-    return cb(null, objectHelper.toCamelCase(data.body || data));
+    return cb(
+      null,
+      objectHelper.toCamelCase(data.body || data, [], { keepOriginal: options.keepOriginalCasing })
```
joshcanhelp: If you're going to keep an options dict here, can't you just pass the whole object in instead of translating?

src/helper/object.js
luisrudge: I'm not passing all the options, it's a new options object and is meant to be internal only (the regular WebAuth options are not passed to this method)

src/helper/response-handler.js
luisrudge: it's a different options object, so no. I can use the same name if you think it's worth it. I used two different names because they're used in different contexts, but I'm fine with both options

src/helper/object.js
joshcanhelp: Understood but does it need a whole options parameter or would a flag work fine here? How much more would you want this function to do?

src/helper/response-handler.js
joshcanhelp: Don't change it on my account 😆 just something to consider. Does it need translation? Does that add cognitive overhead? I won't hold back an approval based on that.

joshcanhelp:
@luisrudge - Also see the review comment:

> A couple of specific comments here but I'm wondering ... is there a strong reason to keep both? Why not just use the `exceptions` array?

luisrudge:
> is there a strong reason to keep both? Why not just use the `exceptions` array?

we have to keep both so we don't introduce a breaking change

src/helper/object.js
luisrudge: it would work today, but an options object makes it future proof

joshcanhelp: It could make it future-overloaded as well ...

luisrudge: I also think using options is better than flags because of this:

`toCamelCase(authResult, ['id_token'], true)`

What does the array mean? What does the true mean? If you don't read the implementation, you'll never know.

PR #933, opened 2019-05-07
Fix telemetry customization

luisrudge:
### Changes

Send default telemetry as `env` when telemetry is customized

### DIFF
--- a/src/helper/request-builder.js
+++ b/src/helper/request-builder.js
@@ -101,7 +101,9 @@ RequestBuilder.prototype.getTelemetryData = function() {
   var telemetryName = this._universalLoginPage ? 'auth0.js-ulp' : 'auth0.js';
   var clientInfo = { name: telemetryName, version: version.raw };
   if (this._telemetryInfo) {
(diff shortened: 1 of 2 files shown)


src/helper/request-builder.js
```diff
@@ -99,7 +99,10 @@ RequestBuilder.prototype.setCommonConfiguration = function(ongoingRequest, optio
 
 RequestBuilder.prototype.getTelemetryData = function() {
   var telemetryName = this._universalLoginPage ? 'auth0.js-ulp' : 'auth0.js';
-  var clientInfo = this._telemetryInfo || { name: telemetryName, version: version.raw };
+  var clientInfo = { name: telemetryName, version: version.raw };
+  if (this._telemetryInfo) {
+    clientInfo = objectHelper.extend({}, this._telemetryInfo, { env: clientInfo });
```
damieng: So this looks like it pushes the standard telemetry info down into the env property... that doesn't sound right to me. @joshcanhelp ?

luisrudge: @damieng that's exactly what was wrong 😬

PR #978, opened 2019-07-18
Upgrade all packages

luisrudge:
### Changes

In order for dependabot to be useful, we have to have up to date packages already, otherwise we'll waste lots of times merging PR by PR. This PR upgrades everything except `superagent` due to its breaking changes. I made separate commits for unit test fixes and also running prettier with the latest version.

### Testing

Sinon had a breaking change as well. Fix is in https://github.com/auth0/auth0.js/commit/0f8164e5d613a5b8f7a1628c2367b8aa5d5d8ae4


luisrudge:
image

PR #1006, opened 2019-08-02
Add missing JSDoc entry for 'username' signup option. Closes #1005

### LINKED ISSUE #1005: [docs] JSDoc form DbConnection.signup does not contain entry for `username`
It seems `username` field of the signup options is not directly documented in the JSDoc, but only in the public docs, which list the field:
https://auth0.com/docs/libraries/auth0js/v9#signup
while the method JSDoc omits it completely:
https://github.com/auth0/auth0.js/blob/cee609cca97eefaa60818bd9b34e16685cf94276/src/authentication/db-connection.js#L21-L3
(I was looking into the api checking something, not into the public docs).
Is there an intent to modify the JSDoc?
Thanks!

peterblazejewicz:
This is a minor correction, aligns JSDoc with online documentation.

Thanks!

### DIFF
--- a/src/authentication/db-connection.js
+++ b/src/authentication/db-connection.js
@@ -24,6 +24,7 @@ function DBConnection(request, options) {
  * @param {Object} options
  * @param {String} options.email user email address
  * @param {String} options.password user password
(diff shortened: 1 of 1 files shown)


luisrudge:
Thanks for the PR 🎉

PR #1040, opened 2019-09-16
also support ionic:// protocol (for iOS ionic web view)

flache:
### Changes

I am using auth0 with cordova. In the cordova app, I am using [`cordova-plugin-ionic-webview`](https://github.com/ionic-team/cordova-plugin-ionic-webview). This plugin serves files on iOS over `ionic://localhost`. It is [not possible](https://github.com/ionic-team/cordova-plugin-ionic-webview#iosscheme) to change the "protocol" into `http(s)://` or `file://`.

Unfortunately, using cross origin authentication in popupMode fails on iOS since the url parser assumes that the protocol is either `http(s)://` or `file://`. When changing the hardcoded regex like in the pull request, everything works as expected. Please consider merging it in for other cordova users.

### DIFF
--- a/src/helper/object.js
+++ b/src/helper/object.js
@@ -124,7 +124,7 @@ function toCamelCase(object, exceptions, options) {
 
 function getLocationFromUrl(href) {
   var match = href.match(
-    /^(https?:|file:)\/\/(([^:/?#]*)(?::([0-9]+))?)([/]{0,1}[^?#]*)(\?[^#]*|)(#.*|)$/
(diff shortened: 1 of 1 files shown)


luisrudge:
Auth0.js is not meant to be used from cordova apps. Check out https://github.com/auth0/auth0-cordova/

flache:
Auth0-Cordova does not support passwordless login with a silent authentication flow and I think it is a quite reasonable user experience request not to require the user to open another app/browser when logging in into our app ¯\_(ツ)_/¯

luisrudge:
There's no Passwordless login support for native applications yet. This is on our roadmap but we're not there yet. For now, you have to use the web flow.

PR #1277, opened 2022-11-14
[IAMRISK-1835] renderPasswordlessChallenge

DominickBattistini:
### Description
Adds new methods to interact with `passwordless/challenge` endpoint when working with Bot Detection in Passwordless flows.

### Reference
https://auth0team.atlassian.net/browse/IAMRISK-1835


src/web-auth/captcha.js
```diff
@@ -201,4 +201,63 @@ function render(auth0Client, element, options, callback) {
...
+ * @param {Object} options The configuration options for the captcha
+ * @param {Object} [options.templates] An object containaing templates for each captcha provider
+ * @param {Function} [options.templates.auth0] template function receiving the challenge and returning an string
+ * @param {Function} [options.templates.recaptcha_v2] template function receiving the challenge and returning an string
+ * @param {Function} [options.templates.recaptcha_enterprise] template function receiving the challenge and returning an string
+ * @param {String} [options.lang=en] the ISO code of the language for recaptcha*
+ * @param {Function} [callback] an optional callback function
+ * @ignore
+ */
+function renderPasswordless(auth0Client, element, options, callback) {
```
robinbijlani, line 219: It looks like `auth0Client` here is the instance of `Authentication`. In a passwordless scenario, would `auth0Client` be an instance of `PasswordlessAuthentication`? (If so, we don't even need a new function here and can re-use `render()` above as-is.

src/web-auth/index.js
```diff
@@ -1123,4 +1123,24 @@ WebAuth.prototype.renderCaptcha = function (element, options, callback) {
...
+ * @param {Object} [options.templates] An object containaing templates for each captcha provider
+ * @param {Function} [options.templates.auth0] template function receiving the challenge and returning an string
+ * @param {Function} [options.templates.recaptcha_v2] template function receiving the challenge and returning an string
+ * @param {Function} [options.templates.recaptcha_enterprise] template function receiving the challenge and returning an string
+ * @param {String} [options.lang=en] the ISO code of the language for recaptcha
+ * @param {Function} [callback] An optional completion callback
+ * @memberof WebAuth.prototype
+ */
+WebAuth.prototype.renderPasswordlessCaptcha = function (element, options, callback) {
+  return captcha.render(this.client, element, options, callback);
```
robinbijlani, line 1143: This isn't using your new `captcha.renderPasswordless()` function, BUT I don't think we need it. If we replace the first argument with `this.client.passwordless`, it looks like it should work.

Would it be reasonable to add something to the `options` we pass in (options.flow, maybe) to the existing `renderCaptcha` function above and swap out which client we use? (The answer may be no. I'm not quite sure what the conventions are here and if a separate function is better.)

src/web-auth/captcha.js
robinbijlani: Edit: It seems like this function is public so we probably do want to create a separate one.

robinbijlani (commented):
Overall this looks good to me. But I am wondering if we need to change any of the docs manually. Maybe the README?

Also, when this goes to DX-SDKs for review, please mention that the new endpoint (`POST /passwordless/challenge`) is not yet merged to server due to the freeze. I am assuming we should hold off on merging this until that change is in.

robinbijlani (dismissed):

evansims (approved):
LGTM 👍

PR #1382, opened 2023-12-14
[IAMRISK-3011] Auth0 V2 Captcha failOpen support

alexkoumarianos-okta:
### Changes
Made changes to allow auth0_v2 to handle failOpen by passing a distinct dummy token.

### References
https://auth0team.atlassian.net/browse/IAMRISK-3011

### DIFF
--- a/src/web-auth/captcha.js
+++ b/src/web-auth/captcha.js
@@ -1,5 +1,5 @@
-// eslint-disable-next-line no-unused-vars
 import Authentication from '../authentication';
+// eslint-disable-next-line no-unused-vars
 import object from '../helper/object';
 
 var noop = function () {};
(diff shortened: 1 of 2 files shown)


src/web-auth/captcha.js
```diff
@@ -189,6 +189,23 @@ function injectCaptchaScript(element, opts, callback, setValue) {
       window.arkose = arkose;
       callback(arkose);
     };
+  } else if (opts.provider === AUTH0_V2_CAPTCHA_PROVIDER) {
+    var a0RetryCount = 0;
```
srijonsaha, line 193: We shouldn't have a separate variable for Auth0 V2, we can use the retryCount variable since when this script runs there will always be one provider so the counter will only represent the retries for that provider.

src/web-auth/captcha.js
```diff
@@ -189,6 +189,23 @@ function injectCaptchaScript(element, opts, callback, setValue) {
...
+    attributes['error-callback'] = function () {
```
srijonsaha, line 194: I don't think this is correct. Are you trying to retry the script loading? That should be done through the attribute 'onerror' like Arkose example above. Also let's try to combine the code since it's identical to Arkose. 

https://developer.mozilla.org/en-US/docs/Web/API/HTMLScriptElement#examples

src/web-auth/captcha.js
```diff
@@ -309,9 +326,23 @@ function handleCaptchaProvider(element, options, challenge) {
...
         if (challenge.provider === AUTH0_V2_CAPTCHA_PROVIDER) {
+          var a0RetryCount = 0;
           renderParams.language = options.lang;
           renderParams.theme = 'light';
+          renderParams.retry = 'never';
+          renderParams['error-callback'] = function () {
+            if (a0RetryCount < MAX_RETRY) {
+              setValue();
+              globalForCaptchaProvider(challenge.provider).reset(widgetId);
+              a0RetryCount++;
```
srijonsaha, line 339: Let's try to use the retryCount variable.

src/web-auth/captcha.js
```diff
@@ -309,9 +326,23 @@ function handleCaptchaProvider(element, options, challenge) {
...
+            } else {
+              // similar implementation to ARKOSE_PROVIDER failOpen
```
srijonsaha, line 341: nit: no need for comment, let's remove

srijonsaha (changes requested):
nice start, left some comments to address

TSLarson (approved):
LGTM

frederikprijck (approved):

PR #1652, opened 2026-05-19
fix(captcha): prevent injection of malicious payloads in challenge.image

amitsingh05667:
## Fix 

The `auth0` captcha template concatenated `challenge.image` directly into an HTML string assigned to `innerHTML`, allowing attribute injection if the value contained quote characters.

**Fix:** removed `challenge.image` from the template string and set `src` via `img.setAttribute('src', challenge.image || '')` instead - the DOM API encodes the value safely.

**No breaking changes.** Public API, visual output, and custom template overrides are all unaffected.

### DIFF
--- a/src/web-auth/captcha.js
+++ b/src/web-auth/captcha.js
@@ -67,20 +67,26 @@ var defaults = {
   }
 };
 
+function escapeAttr(str) {
+  return String(str || '').replace(/&/g, '&amp;').replace(/"/g, '&quot;');
+}
+
 function handleAuth0Provider(element, options, challenge, load) {
-  element.innerHTML = options.templates[challenge.provider](challenge);
-  // Use setAttribute to safely assign challenge.image — avoids HTML injection via innerHTML string concat.
-  // If a custom template is used that omits .captcha-challenge img, src will not be set (by design).
+  var safeChallenge = Object.assign({}, challenge, {
+    image: escapeAttr(challenge.image)
+  });
+  element.innerHTML = options.templates[challenge.provider](safeChallenge);
   var img = element.querySelector('.captcha-challenge img');
   if (img) {
     img.setAttribute('src', challenge.image || '');
   }
-  element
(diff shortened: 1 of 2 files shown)


test/web-auth/captcha.test.js
```diff
@@ -123,6 +123,18 @@ describe('captcha rendering', function () {
       expect(imgEl.src).to.equal(challenges[0].image);
     });
 
+    it('should not inject HTML attributes when challenge.image contains a malicious payload', function () {
```
Piyush-85, line 137: Same test case is duplicated thrice, fix this

src/web-auth/captcha.js
```diff
@@ -71,6 +69,12 @@ var defaults = {
 
 function handleAuth0Provider(element, options, challenge, load) {
   element.innerHTML = options.templates[challenge.provider](challenge);
+  // Use setAttribute to safely assign challenge.image — avoids HTML injection via innerHTML string concat.
+  // If a custom template is used that omits .captcha-challenge img, src will not be set (by design).
+  var img = element.querySelector('.captcha-challenge img');
+  if (img) {
+    img.setAttribute('src', challenge.image || '');
+  }
   element
```
Piyush-85: **Add a gaurd** 
`var reloadBtn = element.querySelector('.captcha-reload');
  if (reloadBtn) {
    reloadBtn.addEventListener('click', function (e) {
      e.preventDefault();
      load();
    });
  }`

src/web-auth/captcha.js
```diff
@@ -71,6 +69,12 @@ var defaults = {
 
 function handleAuth0Provider(element, options, challenge, load) {
   element.innerHTML = options.templates[challenge.provider](challenge);
```
Piyush-85: A custom template gets challenge.image and can still inject it.

Piyush-85 (approved):

PR #1655, opened 2026-05-20
fix(packaging): add `exports` map so default import works under Vite/Vitest

### LINKED ISSUE #1654: v10.0.0: `"type": "module"` + UMD `main` breaks default import under Vite/Vitest (`import auth0 from 'auth0-js'` yields `undefined`)
## Summary

v10.0.0 added `"type": "module"` to `package.json` but kept `main` pointing at the UMD bundle (`dist/auth0.min.js`) and did not add an `exports` map. Bundlers that respect `"type": "module"` (Vite, Vitest, modern esbuild) load the UMD file as ESM and find no top-level `export` statements — so the default import resolves to `undefined`, breaking the documented `new auth0.WebAuth(...)` / `new auth0.Authentication(...)` usage shown in the README.

## Reproduction

```sh
mkdir auth0-v10-repro && cd auth0-v10-repro
npm init -y
npm i auth0-js@10.0.0 vite vitest
cat > probe.test.js <<'EOF'
import auth0 from 'auth0-js';
import * as ns from 'auth0-js';

test('default + namespace shape', () => {
  console.log('default import:', auth0);                 // undefined
  console.log('namespace keys :', Object.keys(ns));      // []
});
EOF
npx vitest run probe.test.js
```

Output:

```
default import: undefined
namespace keys: []
```

Importing the ESM bundle directly works as expected:

```js
import auth0 from 'auth0-js/dist/auth0.min.esm.js';
// → { Authentication: [Function], Management: [Function], WebAuth: [Function], version: '10.0.0' }
```

Jest is unaffected because its resolver loads `main` as CJS and the UMD wrapper's `exports.X = ...` writes work correctly under CJS execution.

## Why this happens

`package.json` in v10:

```json
{
  "type": "module",
  "main": "dist/auth0.min.js",
  "module": "dist/auth0.min.esm.js"
}
```

- **Node ESM / Vite / Vitest / modern esbuild:** sees `"type": "module"`, loads the resolved file (`main` in many configs) as ESM. The UMD's IIFE writes to `module.exports` / `exports.X` at runtime, which produces no ESM exports → empty namespace, undefined default.
- **Jest:** falls back to CJS resolution, runs `main` as CJS, UMD writes succeed.

The `module` field is supposed to win for ESM-aware bundlers, but in practice Vite's resolution interacts with `"type": "module"` such that the UMD `main` gets pulled in as ESM. The `exports` field is the modern, authoritative way to disambiguate.

## Impact

Every consumer using Vite, Vitest, or recent esbuild (i.e., the default tooling for new React/Vue/Svelte/Solid apps since ~2023) breaks on `v9 → v10` upgrade. The failure surfaces as a runtime `TypeError: Cannot read properties of undefined (reading 'Authentication')` (or `'WebAuth'`) at whatever call site first uses the SDK constructors — typically inside `componentDidMount` / route initialization.

The README itself documents the broken usage:

> ```js
> import auth0 from 'auth0-js';
> var webAuth = new auth0.WebAuth({ ... });
> ```

## Suggested fix

Add a conditional `exports` map so each environment receives the right bundle:

```json
{
  "type": "module",
  "main": "dist/auth0.min.js",
  "module": "dist/auth0.min.esm.js",
  "exports": {
    ".": {
      "import": "./dist/auth0.min.esm.js",
      "require": "./dist/auth0.min.js",
      "default": "./dist/auth0.min.esm.js"
    }
  }
}
```

Alternatives:
- Point `main` at the ESM bundle (simpler, but breaks any consumer that's required CJS via `require('auth0-js')`).
- Drop `"type": "module"` (regresses the v10 intent of marking the package as ESM).

Either of these restores the documented `import auth0 from 'auth0-js'` usage across all bundlers.

## Workarounds for affected consumers

1. **Pin to `^9.29.0`** until v10 packaging is fixed.
2. **Vite/Vitest only:** alias `auth0-js` to its ESM bundle in your bundler config:
   ```js
   resolve: { alias: { 'auth0-js': 'auth0-js/dist/auth0.min.esm.js' } }
   ```

## Environment

- `auth0-js@10.0.0`
- Vite 7.x, Vitest 4.x
- Node 22.x and 24.x (same behavior)
- Bundler config: defaults from `@vitejs/plugin-react`

Happy to send a PR with the `exports` map if maintainers prefer that path.

MatthewHarrigan:
Fixes #1654.

### Changes

Adds an `exports` field to `package.json` so each module environment resolves `auth0-js` to the correct bundle. No code changes; no `dist/` rebuild; no build/test infrastructure changes.

```json
"exports": {
  ".": {
    "import":  "./dist/auth0.min.esm.js",
    "require": "./dist/auth0.min.js",
    "default": "./dist/auth0.min.esm.js"
  },
  "./dist/*":        "./dist/*",
  "./build/*":       "./build/*",
  "./package.json":  "./package.json"
}
```

**Why this works:**

- `import` condition → ESM bundle. Fixes the bug reported in #1654: Vite/Vitest (and any modern bundler that respects `"type": "module"`) were resolving the package to the UMD `main`, loading it as ESM, finding no top-level `export` statements, and handing consumers `undefined` for the default import. Routing `import` directly at the ESM bundle restores the documented `import auth0 from 'auth0-js'` usage.
- `require` condition → UMD bundle. Jest-style transformers that read `exports` will pick up the UMD file and execute it as CJS, where `exports.X = ...` writes work correctly.
- `default` fallback → ESM bundle for any environment that doesn't match the above conditions.
- `./dist/*` and `./build/*` wildcards preserve deep-path imports (`auth0-js/dist/auth0.min.esm.js` etc.) that some consumers have adopted as a workaround for #1654. Once this lands the workaround is unnecessary, but keeping the wildcards avoids breaking anyone who's still on the workaround.
- `./package.json` is exposed because some tools (TypeScript, bundlers) read it.

### Honest scope

This PR fixes the **ESM consumer path** (the bug in #1654 — the one breaking the documented import). The pure-Node **CJS `require()` path** remains broken on v10 because `"type": "module"` causes Node to load `.js` files as ESM regardless of the `require` condition target, so the UMD bundle's `exports.X = ...` writes still have no effect. Fixing that properly requires renaming the UMD output to `.cjs` (or splitting the build), which is out of scope here. Filing as a separate concern if maintainers want.

Verified on a baseline of v10.0.0 master (pre-PR): `require('auth0-js')` already returns `[Module: null prototype] {}` — so this PR neither fixes nor regresses CJS.

### References

- Bug report with full reproduction: #1654
- README documents `import auth0 from 'auth0-js'` then `new auth0.WebAuth({...})` — the exact pattern broken by v10 packaging.

### Testing

All existing local CI gates green against this branch:

| Gate | Result |
|---|---|
| `npm run lint` | clean |
| `npm run build` | produces all 4 expected artifacts |
| `npm test` | 658 passing |
| `npm run test:es-check:es5` | pass |
| `npm run test:es-check:es2015:module` | pass |

I didn't run `npm run smoke:modern` or the BrowserStack e2e tests — those need toolchain I don't have set up locally. Happy to add coverage if the reviewer suggests a specific gate.

#### Resolution probe (Node 22, ESM consumer)

Against the unmodified v10.0.0:
```
default import: undefined
namespace keys: []
```

Against this branch:
```
default import: { Authentication: [Function], Management: [Function], WebAuth: [Function], version: '10.0.0' }
default.Authentication: function
default.WebAuth: function
```

- [ ] This change adds unit test coverage — *N/A, packaging-only metadata change*
- [ ] This change adds integration test coverage — *N/A, packaging-only metadata change*

### Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,16 @@
   "type": "module",
   "module": "dist/auth0.min.esm.js",
   "main": "dist/auth0.min.js",
+  "exports": {
+    ".": {
+      "import": "./dist/auth0.min.esm.js",
+      "require": "./dist/auth0.min.js",
+      "default": "./dist/auth0.min.esm.js"
+    },
+    "./dist/*": "./dist/*",
+    "./build/*": "./build/*",
+    "./package.json": "./package.json"
+  },
   "files": [
     "src",
     "plugins",


MatthewHarrigan:
Hi @amitsingh05667 - I've updated the branch

MatthewHarrigan:
Hi @amitsingh05667 - is it possible to take another look at this if you have time?

cschetan77:
Hi @MatthewHarrigan, apologies for the late response.
Thanks for fixing this. As you pointed out, the `require()` path remains broken, the require condition points at `auth0.min.js`, and with `"type": "module"` set, Node refuses to load a `.js` file as CJS. That can be addressed separately. 
Looks good to me. 
I am going ahead and merging it, will release a patch shortly.

cschetan77 (approved):

cschetan77:
@MatthewHarrigan I think there is one commit that doesn't have a verified signature. Could you please sign all the commits and re push them.

MatthewHarrigan:
Hi @cschetan77 - I have tidied up the commits - please can you review

cschetan77:
Thanks @MatthewHarrigan , Could you rebase your branch as well.

MatthewHarrigan:
Hi @cschetan77 - can we merge this in now?

cschetan77 (approved):

